PR #23, opened 2019-12-25
Add Auth Mount Path Config (Env / Annotation)

shlee322:
### Method 1. Set sidecar-injector container environment variable
```yaml
- name: AGENT_INJECT_VAULT_AUTH_PATH
  value: "auth/kubernetes"
```

### Method 2. Add target pod annotation
```yaml
vault.hashicorp.com/auth-path: "auth/kubernetes"
```

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -15,6 +15,7 @@ import (
 
 const (
 	DefaultVaultImage = "vault:1.3.1"
+	DefaultVaultAuthPath = "auth/kubernetes"
 )
 
 // Agent is the top level structure holding all the
@@ -101,6 +102,9 @@ type Vault struct {
 	// Address is the Vault service address.
 	Address string
 
+	// AuthPath is the Mount Path of Vault Kubernetes Auth Method.
+	AuthPath string
+
 	// CACert is the name of the Certificate Authority certificate
 	// to use when validating Vault's server certificates.
 	CACert string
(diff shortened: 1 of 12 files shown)


joemiller:
We would also like to see the auth mount path configurable. However, it would be preferable (in our case) for it to be configurable in the agent-inject webhook

Use case:
- Vault instance serving multiple kubernetes clusters.
- Each k8s cluster is registered at a different mount path, eg:
1. `/auth/k8s-cluster01`
2. `/auth/k8s-cluster02`
3. `/auth/k8s-cluster03`
- Applications do not know what cluster they're deployed into, and may be deployed to multiple clusters.
- These apps' deployment yaml's should not need any cluster-specific information in them, so specifying the auth mount path in the pod's annotations would not be ideal here.
- The agent-inject webhook (ideally) would be deployed into each cluster with a config option specifying the auth path relevant to that cluster, eg: `https://$VAULT_ADDR` + `/auth/k8s-cluster01`
- The agent config file generated by the agent-injector would set the auth_auth section with the correct mount path for the current k8s cluster

joemiller:
_(deleted previous comment, as i misread this as a feature request not an open PR, apologies)_

Would love to see this merged! Thank you for this, we have this same use case / need.

carnei-ro:
The same use case here! Thanks for implementing it.

izhukov:
Also hit the lack of this feature. We use the external vault with several kubernetes auth backends.
@shlee322 much appreciated your effort

devorbitus:
This feature would really help us when it is merged

jesseaukeman:
Lack of this feature is also blocking our use.  I assume this environment variable will be added as part of the  vault-helm chart and the default configurable in the chart values.yaml?

Mark-McCracken:
Now that this is merged, can someone perform a release to dockerhub to update this so we can use it without having to build ourselves?

Mark-McCracken:
I tested this last night, works like a charm with multiple clusters :)
Thanks so much for this @shlee322 that's so much easier than having to use a custom configmap!

jasonodonnell:
Released in `v0.2.0`.  Hope that helps!

Mark-McCracken:
Thanks so much :)

Issue #33, opened 2020-01-08
Agent skips TLS verification by default


xntrik:
As per 3rd party report
https://github.com/hashicorp/vault-k8s/blob/8f3a6118fb8b640e9439523f758703973dbdd472/agent-inject/agent/annotations.go#L228

jasonodonnell:
The fix has been merged, test coverage was added to ensure this stays fixed and `vault-k8s:0.1.2` was released.

xntrik:
Thanks to @fischerman for reporting the issue!

PR #34, opened 2020-01-08
TLS verification should not be skipped by default

xntrik:
As per https://github.com/hashicorp/vault-k8s/issues/33


jasonodonnell:
Thanks @xntrik!

Issue #56, opened 2020-01-23
add support for setting "command" on templates


joemiller:
Use case: Need to run a command (eg: send `SIGHUP`) when secrets are updated.

A useful example is nginx and fetching mTLS certs from Vault. After the cert is updated we need to send a `SIGHUP` to trigger nginx to reload the certs.

It looks like this is currently possible using the configmap escape hatch but this is a common enough case for us that having to use the configmap approach is significant friction.

Example of how this might look as annotations:

```yaml
        vault.hashicorp.com/agent-inject-secret-tls.pem: "pki/issue/ou-nobody"
        vault.hashicorp.com/agent-inject-template-tls.pem: |
           {{- with secret "pki/issue/ou-nobody" "common_name=foo.example.com" "ttl=5m" }}
           {{ .Data.private_key }}
           {{ .Data.certificate }}
           {{ .Data.issuing_ca }}
           {{ end }}
        vault.hashicorp.com/agent-inject-command-tls.pem: "/bin/sh -c 'pkill -HUP nginx || true'"

  spec:
    # shared PID namespace so the vault-agent sidecar can send signals to nginx in the app container
    shareProcessNamespace: true
```

PR #57, opened 2020-01-23
add "command" annotation for executing after secret template is rendered

joemiller:
Addresses #56 

This PR adds a new annotation:
* `vault.hashicorp.com/agent-inject-command-` 

This annotation similar to the one used for customizing the secret template. The specified command will be run by vault-agent after the secret is rendered.

Unaddressed by this PR is that in many cases the vault agent will be running as a different user than the main app container. Thus, the example use case from #56 is not yet fully possible since the call to `pkill` will be run as a different user/uid than the nginx container.

It may be desirable to allow an annotation that will set the UID of the vault agent container to address this.

**EDIT**: support for specifying the userid/groupid of the vault agent sidecar is addressed in #60

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -39,6 +39,13 @@ const (
 	// from auth/token/lookup-self
 	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token"
 
+	// AnnotationAgentInjectCommand is the key annotation that configures Vault Agent
+	// to run a command after the secret is rendered. The name of the template is any
+	// unique string after "vault.hashicorp.com/agent-inject-command-". This should map
+	// to the same unique value provided in ""vault.hashicorp.com/agent-inject-secret-".
+	// If not provided (the default), no command is executed.
+	AnnotationAgentInjectCommand = "vault.hashicorp.com/agent-inject-command"
+
 	// AnnotationAgentImage is the name of the Vault docker image to use.
 	AnnotationAgentImage = "vault.hashicorp.com/agent-image"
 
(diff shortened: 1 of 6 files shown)


jasonodonnell:
Hi @joemiller, my apologies for the delay.  I'm ready to review this PR.  Can you resolve the conflicts in the meantime? Thanks!

joemiller:
Will do. I’m traveling so it may be a few days

joemiller:
@jasonodonnell conflicts resolved. The conflict in annotations_test.go was a bit of a challenge to resolve but I believe it's in the correct state now. Please advise if I missed something there. thanks!

agent-inject/agent/config.go
```diff
@@ -70,6 +70,7 @@ type Template struct {
 	Contents       string `json:"contents"`
 	LeftDelim      string `json:"left_delimiter,omitempty"`
 	RightDelim     string `json:"right_delimiter,omitempty"`
+	Command        string `json:"command",omitempty`
```
jasonodonnell: The closing quote needs to be at the end of the line before the tick.

agent-inject/agent/annotations.go
```diff
@@ -38,6 +38,12 @@ const (
 	// AnnotationAgentInjectToken is the annotation key for injecting the token
 	// from auth/token/lookup-self
 	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token"
+	// AnnotationAgentInjectCommand is the key annotation that configures Vault Agent
```
jasonodonnell: Nit: can you add a newline before this comment?

joemiller:
@jasonodonnell great eye! thanks. Fixed, pushed

sbeaulie:
Are you planning to get a release for this soon?

Shaeli:
Hello,
Any news on when this will be released? 
Thanks

tvoran:
@Shaeli @sbeaulie It should go out this week, in #99

joemiller:
It would be great to get it released together with the run-as-user/group
support in https://github.com/hashicorp/vault-k8s/pull/60 since your
command may need to be run as a user/group other than the vault defaults
(100/1000) but that PR is still WIP (hopefully done soon)

PR #60, opened 2020-01-25
allow agent container RunAsUser/Group to be specified via annotations

joemiller:
This PR adds two new annotations:

* `vault.hashicorp.com/agent-run-as-user`
* `vault.hashicorp.com/agent-run-as-group`

Which allows overriding the default userid (100) and groupid (1000) that the Vault Agent init and sidecar containers will run as.

This complements the `vault.hashicorp.com/agent-inject-command-` annotation in https://github.com/hashicorp/vault-k8s/pull/57

An example use case for these PRs is described in https://github.com/hashicorp/vault-k8s/issues/56

TL;DR -- Given an example such as running an nginx container with dynamic TLS certs retrieved from Vault, it is necessary to send a `SIGHUP` to nginx to load the new certs. This requires specifying a command to run after new certs are rendered (#57) and that command will need to be run from the Vault Agent sidecar as the same userID that nginx is running in the main/app container (addressed in this PR).


joemiller:
Yeah, makes sense. Will do

joemiller:
resolved merge conflicts.
still TODO: env vars

joemiller:
@jasonodonnell I'm looking at the env var code in `subcommand/injector/flags.go` and I see that most (all?) env vars are also configurable via `-flags`. Just to confirm, we would want to implement both here too?

jasonodonnell:
@joemiller Yessir.  It's a duplication of functionality but flags existed before envs, but envs are a bit easier to configure in Kube vs running the binary.

joemiller:
@jasonodonnell I'm following the pattern used to propagate `AGENT_INJECT_VAULT_IMAGE` env var thru the codebase and I think I'll need to increase the arity on the `agent.Init()` func to include two new vars - user and group:

https://github.com/hashicorp/vault-k8s/blob/master/agent-inject/agent/annotations.go#L137

Are you ok if I proceed this way? Or refactor? perhaps modifying `agent.Init()` to take a struct of default annotations/options instead?

jasonodonnell:
@joemiller Yeah I'd love that actually, it's becoming quite big as is.

joemiller:
@jasonodonnell Updated to allow setting user/group at a global level via env vars: `AGENT_INJECT_RUN_AS_{USER,GROUP}`.

I refactored the code from the original commit quite a bit for this. I chose to represent the user/group as strings for the purpose of -flags and env vars, ultimately converting to int64's just before usage in the Agent. This simplified a lot of the testing and parsing.

Also modified the `agent.Init()` func to take an `AgentConfig{}` struct per discussion above in order to avoid increasing the arity of the func.

It looks like I have quite a few files to resolve conflicts before it is mergeable but the code itself should be ready for review now.

joemiller:
yeah, quite a lot of conflicts to sort out :(

joemiller:
Ok, I think this is ready now. There were a significant amount of conflicts to resolve from the merge of the revokeOnShutdown work that touched all of the same parts of code as this PR (lesson learned: be fast ;) )

jasonodonnell:
Hi @joemiller, just took a look at this.  The code looks OK but I'm having issues with functional testing.

First, when specifying a non-vault user, the following appears when trying to sink the token:

```
Error creating file sink: error during write check: error opening temp file in dir /home/vault for writing: open /home/vault/.vault-token.tmp.b405fff6: permission denied
```

Next, although I set a custom group, only `runAsUser` appears (`runAsGroup` has no value).

Here's a complicated example I am using to test this:

```yaml
spec:
  template:
    metadata:
      annotations:
        # AGENT INJECTOR SETTINGS
        vault.hashicorp.com/agent-run-as-user: "2000"
        vault.hashicorp.com/agent-run-as-group: "100"
        vault.hashicorp.com/secret-volume-path: "/etc/test"
        vault.hashicorp.com/agent-inject-status: "update"
        # DATABASE SECRET
        vault.hashicorp.com/secret-volume-path-db-creds: "/vault/secretfoo"
        vault.hashicorp.com/agent-inject-secret-db-creds: "database/creds/db-app"
        vault.hashicorp.com/agent-inject-template-db-creds: |
          {{- with secret "database/creds/db-app" -}}
          postgres://{{ .Data.username }}:{{ .Data.password }}@postgres.postgres.svc:5432/wizard?sslmode=disable
          {{- end }}
        # TLS SERVER CERTIFICATE
        vault.hashicorp.com/agent-inject-secret-server.cert: "pki/issue/hashicorp-com"
        vault.hashicorp.com/agent-inject-template-server.cert: |
          {{- with secret "pki/issue/hashicorp-com" "common_name=www.hashicorp.com" -}}
          {{ .Data.certificate }}
          {{- end }}
        # TLS SERVER KEY
        vault.hashicorp.com/agent-inject-secret-server.key: "pki/issue/hashicorp-com"
        vault.hashicorp.com/agent-inject-template-server.key: |
          {{- with secret "pki/issue/hashicorp-com" "common_name=www.hashicorp.com" -}}
          {{ .Data.private_key }}
          {{- end }}
        # TLS CA CERTIFICATE
        vault.hashicorp.com/agent-inject-secret-ca.cert: "pki/issue/hashicorp-com"
        vault.hashicorp.com/agent-inject-template-ca.cert: |
          {{- with secret "pki/issue/hashicorp-com" "common_name=www.hashicorp.com" -}}
          {{ .Data.issuing_ca }}
          {{- end }}
        # VAULT SETTINGS
        vault.hashicorp.com/role: "db-app"
        vault.hashicorp.com/tls-secret: "tls-test-client"
        vault.hashicorp.com/client-cert: "/vault/tls/client.crt"
        vault.hashicorp.com/client-key: "/vault/tls/client.key"
        vault.hashicorp.com/ca-cert: "/vault/tls/ca.crt"
          {{- with secret "auth/token/lookup-self" -}}{{.Data.id}}{{- end }}
```

joemiller:
@jasonodonnell I'll look at this today, as soon as I can.

joemiller:
@jasonodonnell Ok, I can re-create the issue with being unable to write the token when the user is neither '0' nor the default vault user in the vault docker image.

I'm going to look at seeing if it is possible to set the ownership on the /home/vault mount to the same user as is specified by run-as-{user,group}

I am also seeing the 'group' GID is not being set in the container:
```
# vault-injector (webhook) deploy:
         - name: AGENT_INJECT_RUN_AS_USER
           value: "0"
         - name: AGENT_INJECT_RUN_AS_GROUP
           value: "1000"
```
```
# logs from vault-agent pod:
2020/03/02 23:05:49.751264 [INFO] (child) spawning: /bin/sh -c pkill -HUP nginx || true; whoami; id
root
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

joemiller:
RE: group not being set. I am wondering if this is a behavior in k8s. I'm using the example pod from: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
which specifies:
```yaml
  securityContext:
    runAsUser: 1000
    runAsGroup: 3000
    fsGroup: 2000
```
And the observed outcome is different than what is illustrated in the docs:
```console 
$ ku exec -it security-context-demo sh                             
/ $ id
uid=1000 gid=0(root) groups=2000
```

joemiller:
Ah, I think I see -- `runAsGroup` is feature-gated in 1.13, full available in 1.14. I'm testing against a 1.13 (GKE) cluster and don't have a 1.14 cluster readily available for testing. @jasonodonnell are you testing against 1.13 too?  

https://github.com/rancher/rancher/issues/19901#issuecomment-488632314

joemiller:
And I think we can fix the permissions issue by setting `fsGroup` to the value of `runAsGroup`, however `fsGroup` is only available in the `PodSecurityContext{}` and not the container `SecurityContext{}` that we're currently setting so this will require a bit more code to modify the pod's securitycontext

joemiller:
Ok, another update.

Perhaps due to my misunderstanding, but now I'm starting to grok things a little more.

So, both the init container and the sidecar container store the token in `/home/vault` which is not shared between the pods. Since this dir is not a volume mount, the pod `fsGroup` setting doesn't help here.

I'm wondering if we should do the following:

1. Only set User/Group/fsGroup on the **sidecar** container, not on the init container.
2. This means we should ignore the `vault.hashicorp.com/agent-inject-command-` annotation when configuring the init container's vault config. We would do this because presumably the command is not going to run as the user/group specified and is likely to fail.
3. We would also ignore runas-user/group on the init container and it will always run as the default uid/gid 100/1000

jasonodonnell:
@joemiller 

> Ah, I think I see -- runAsGroup is feature-gated in 1.13, full available in 1.14. I'm testing against a 1.13 (GKE) cluster and don't have a 1.14 cluster readily available for testing. @jasonodonnell are you testing against 1.13 too?

Indeed I was, this makes sense!

> And I think we can fix the permissions issue by setting fsGroup to the value of runAsGroup, however fsGroup is only available in the PodSecurityContext{} and not the container SecurityContext{} that we're currently setting so this will require a bit more code to modify the pod's securitycontext

We likely shouldn't go mucking with pod fsGroup as it could cause issues with other containers in the pod.

Here's what I think should work: we can make `/home/vault` an `emptyDir` volume.  This should get around issues with permissions.  Thoughts?

joemiller:
I’ll try that

joemiller:
@jasonodonnell After a quick local POC I think the emptyDir volume for `/home/vault` will work. I need to plish up the code and update tests. I'll work on that today, though I have a busy day so might be a little delayed.

In summary, it should look like:

- In both the `init` container and `sidecar` container:
- set the container's securityContext user/group based on either: Defaults, annotations (run-as-user/group), or env vars
- Both `init` and `sidecar` will share an emptyDir `/home/vault` where they will write the tokens. This is a slight change in behavior since the current design has both containers with their own `/home/vault` dirs. I think this is just fine. In theory I suppose vault-agent in the sidecar could re-use the token from the init container although I don't think vault-agent looks for an existing token in the sink file at startup.

joemiller:
@jasonodonnell Pushed a new (separate) commit implementing the emptyDir for `/home/vault` - https://github.com/hashicorp/vault-k8s/pull/60/commits/4335a2318f7a6bfe866dda9ff20c41054d72f782

Made it a separate commit for review purposes. It could be squashed.

I chose 'token volume' as the name for variables and func names. Could change it to 'home volume' tho.

joemiller:
@jasonodonnell Rebased to address merge conflicts but I'm stuck on this new test failure:

```
go test -race ./...
?       github.com/hashicorp/vault-k8s  [no test files]
ok      github.com/hashicorp/vault-k8s/agent-inject     (cached)
--- FAIL: TestTemplateShortcuts (0.00s)
    --- FAIL: TestTemplateShortcuts/valid_inject-token (0.00s)
        annotations_test.go:342: expected secret {Name:token Path:auth/token/lookup-self Template:{{ with secret "auth/token/lookup-self" }}{{ .Data.id }}
            {{ end }} MountPath: Command:}, got agent secret {Name:token Path:auth/token/lookup-self Template:{{ with secret "auth/token/lookup-self" }}{{ .Data.id }}
            {{ end }} MountPath:/vault/secrets Command:}
FAIL
FAIL    github.com/hashicorp/vault-k8s/agent-inject/agent       0.219s
ok      github.com/hashicorp/vault-k8s/helper/cert      (cached)
ok      github.com/hashicorp/vault-k8s/subcommand/injector      (cached)
?       github.com/hashicorp/vault-k8s/subcommand/version       [no test files]
?       github.com/hashicorp/vault-k8s/version  [no test files]
FAIL
make: *** [unit-test] Error 1
```

jasonodonnell:
Sorry we couldn't get this into v0.3.0, but this will definitely be in the next release!

joemiller:
@jasonodonnell Awesome. I rebased to bring in the latest master (agent.go, vaultImage tag update)

joemiller:
@jasonodonnell I looked around to see if there were any docs for the injector that I should try to update to include these annotations but I couldn't find anything. I'm happy to send PRs to document this and #57 if that is available.

jasonodonnell:
@joemiller Thanks Joe, I can update them once this is merged.  I plan to review this tomorrow!

sbeaulie:
What's the update on this? I came accross the need for it for secrets populated in a `.ssh` directory since SSH has strict requirements for the file ownership and permissions. The files need to be user owned, and the directory to be 700 perms and files 600. The perms are supported via annotations, but not the owner uid/gid yet.

PR #61, opened 2020-01-28
turn on GO111MODULE in build

Kaizhe:
Signed-off-by: kaizhe <derek0405@gmail.com>

turn on `GO111MODULE` in case it is not set in people's env.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ BIN_NAME=$(IMAGE_NAME)_$(GOOS)_$(GOARCH)_$(VERSION)
 all: build
 
 build:
+	export GO111MODULE=on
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,6 +13,7 @@ BIN_NAME=$(IMAGE_NAME)_$(GOOS)_$(GOARCH)_$(VERSION)
 all: build
 
 build:
+	export GO111MODULE=on
```
jasonodonnell: Is `export` really needed or could this be prepended to the `go build` command like the other envs?

Kaizhe: Removed, thanks!

PR #71, opened 2020-02-06
Allow overriding of volume mount path and added a flag to preserve case sensitivity of secrets

smitthakkar96:
This PR adds two new annotations:
- `vault.hashicorp.com/secret-volume-path`
- `vault.hashicorp.com/preserve-secret-case`
- Fixes broken test (it was always passing): [TestSecretAnnotations](https://github.com/hashicorp/vault-k8s/blob/master/agent-inject/agent/annotations_test.go#L107)

Why?
- Currently vault-injector doesn't allow overriding the path where it mounts the secrets. In our case our base image expects the secrets to be present in a particular directory so it gets picked up as env vars when the container starts. 

- Currently vault-injector assumes that the case of the secret name doesn't matter well in our case it is very important, so to not break backwards compatibility and also fullfil the need I added `vault.hashicorp.com/preserve-secret-case`


agent-inject/agent/container_volume.go
```diff
@@ -8,7 +8,6 @@ const (
 	configVolumeName    = "vault-config"
 	configVolumePath    = "/vault/configs"
 	secretVolumeName    = "vault-secrets"
-	secretVolumePath    = "/vault/secrets"
```
jasonodonnell: This constant should remain and used if `AnnotationVaultSecretVolumePath` is not set.

agent-inject/agent/annotations.go
```diff
@@ -112,6 +112,12 @@ const (
 	// AnnotationVaultAuthPath specifies the mount path to be used for the Kubernetes auto-auth
 	// method.
 	AnnotationVaultAuthPath = "vault.hashicorp.com/auth-path"
+
+	// AnnotationVaultSecretVolumePath specifies where the secrets are to be
+	// Mounted after fetching.
+	AnnotationVaultSecretVolumePath = "vault.hashicorp.com/secret-volume-path"
+
+	AnnotationPreserveSecretCase = "vault.hashicorp.com/preserve-secret-case"
```
jasonodonnell: Needs comment for go doc.

Makefile
```diff
@@ -1,6 +1,6 @@
 REGISTRY_NAME?=docker.io/hashicorp
 IMAGE_NAME=vault-k8s
-VERSION?=0.2.0
+VERSION?=0.2.1
```
jasonodonnell: Please change this back to `0.2.0`, we'll handle release increments.

agent-inject/agent/annotations.go
```diff
@@ -173,6 +179,13 @@ func Init(pod *corev1.Pod, image, address, authPath, namespace string) error {
 		pod.ObjectMeta.Annotations[AnnotationAgentRequestsMem] = DefaultResourceRequestMem
 	}
 
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationPreserveSecretCase]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationPreserveSecretCase] = "false"
+	}
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationVaultSecretVolumePath]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationVaultSecretVolumePath] = "/vault/secrets"
```
jasonodonnell: This should use the secret path constant instead.

agent-inject/agent/annotations.go
```diff
@@ -173,6 +179,13 @@ func Init(pod *corev1.Pod, image, address, authPath, namespace string) error {
 		pod.ObjectMeta.Annotations[AnnotationAgentRequestsMem] = DefaultResourceRequestMem
 	}
 
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationPreserveSecretCase]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationPreserveSecretCase] = "false"
```
jasonodonnell: This should be a bool to be consistent with other annotations: https://github.com/hashicorp/vault-k8s/blob/master/agent-inject/agent/annotations.go#L210-L244

agent-inject/agent/annotations.go
```diff
@@ -188,7 +201,12 @@ func secrets(annotations map[string]string) []*Secret {
 		secretName := fmt.Sprintf("%s-", AnnotationAgentInjectSecret)
 		if strings.Contains(name, secretName) {
 			raw := strings.ReplaceAll(name, secretName, "")
-			name := strings.ToLower(raw)
+			var name string
+			if annotations[AnnotationPreserveSecretCase] == "false" {
```
jasonodonnell: Should use boolean value here.

smitthakkar96:
Thanks @jasonodonnell will address the comments

smitthakkar96:
@jasonodonnell I addressed the comments

smitthakkar96:
@jasonodonnell any updates?

ShabihZaidi:
any update, I want this feature too

jasonodonnell:
Hi @smitthakkar96, sorry for the delay, but I'm ready to review this now!  When you get a chance can you resolve the conflicts?  Thanks!

smitthakkar96:
@jasonodonnell resolved

agent-inject/agent/annotations_test.go
```diff
@@ -135,20 +135,60 @@ func TestSecretAnnotations(t *testing.T) {
...
-				}
 			}
 		} else if len(agent.Secrets) > 0 {
 			t.Error("Secrets length was greater than zero, it shouldn't have been")
 		}
+	}
+}
+
+func TestSecretAnnotationsWithPreserveCaseSensitivityFlagOn(t *testing.T) {
+	tests := []struct {
```
smitthakkar96: @jasonodonnell this is not dry but making it dry would make tests a bit complicated so I duplicated the code

smitthakkar96:
ping!

smitthakkar96:
@jasonodonnell isn't it better to still keep it global and allow overriding at secret level if needed? I am happy to update the code to allow overriding the path and case sensitivity flag at secret level if we agree

saurav-k:
@smitthakkar96 @jasonodonnell   Your comments make more sense to have path  configurable per secret. Specially with pattern you suggested .

smitthakkar96:
@saurav-k I still think having a global flag makes it easy for people to configure incase if you want to override the path for all secrets you are fetching. Incase of if you want to override per secret you can still do if we implement what @jasonodonnell is suggesting

saurav-k:
@smitthakkar96  Makes sense to have a global flag in case you want to mount all secret at same path and  this PR can be scoped to global flag and another PR can be created for the feature suggested by @jasonodonnell   , My thoughts on suggested feature. I have seen apps need files from different mount location as secret  . and in that case suggested feature will be very useful. Let's wait for @jasonodonnell  to give his view on scope of this PR.

jasonodonnell:
@smitthakkar96 I agree the global flag is also valuable because in some situations it would eliminate a lot of annotations. It would likely be that secret specific configurations override the globals.

smitthakkar96:
Ok I will work on it tonight and hopefully update the PR.

Best Regards,
Smit

smitthakkar96:
@jasonodonnell I made the changes, please have a look when you get chance

smitthakkar96:
@jasonodonnell thanks for merging it :)

PR #77, opened 2020-02-14
Support for a "token-only" injection annotation

### LINKED ISSUE #16: Add support for token only
There are use cases where a valid token could be shared with other containers in the pod.  I think an annotation could be added to support this use case.

```yaml
vault.hashicorp.com/agent-inject-token-only: "true"
vault.hashicorp.com/role: "myrole"
```
vs

```yaml
    {{- with secret "auth/token/lookup-self" -}}
        {{.Data.id}}
     {{- end }}
 vault.hashicorp.com/role: "myrole"
```

tvoran:
The annotation `"vault.hashicorp.com/agent-inject-token": "true"`
results in a `token` file containing the `lookup-self` token.

Resolves #16

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -35,6 +35,10 @@ const (
 	// If not provided, a default generic template is used.
 	AnnotationAgentInjectTemplate = "vault.hashicorp.com/agent-inject-template"
 
+	// AnnotationAgentInjectToken is the annotation key for injecting the token
+	// from auth/token/lookup-self
+	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token-only"
+
 	// AnnotationAgentImage is the name of the Vault docker image to use.
 	AnnotationAgentImage = "vault.hashicorp.com/agent-image"
 
@@ -184,6 +188,11 @@ func Init(pod *corev1.Pod, image, address, authPath, namespace string) error {
 // name: foobar, value: db/creds/foobar
 func secrets(annotations map[string]string) []*Secret {
 	var secrets []*Secret
+	// First check for the token-only injection annotation
+	if _, found := annotations[AnnotationAgentInjectToken]; found {
+		annotations[fmt.Sprintf("%s-%s", AnnotationAgentInjectSecret, "token")] = TokenSecret
+		annotations[fmt.Sprintf("%s-%s", AnnotationAgentInjectTemplate, "token")] = TokenTemplate
+	}
 	for name, path := range annotations {
 		secretName := fmt.Sprintf("%s-", AnnotationAgentInjectSecret)
 		if strings.Contains(name, secretName) {
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -37,7 +37,7 @@ const (
 
 	// AnnotationAgentInjectToken is the annotation key for injecting the token
 	// from auth/token/lookup-self
-	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token-only"
+	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token"
 
 	// AnnotationAgentImage is the name of the Vault docker image to use.
 	AnnotationAgentImage = "vault.hashicorp.com/agent-image"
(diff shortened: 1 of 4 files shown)


agent-inject/agent/config.go
```diff
@@ -8,6 +8,8 @@ import (
 
 const (
 	DefaultTemplate = "{{ with secret \"%s\" }}{{ range $k, $v := .Data }}{{ $k }}: {{ $v }}\n{{ end }}{{ end }}"
+	TokenTemplate   = `{{ with secret "auth/token/lookup-self" }}{{ .Data.id }}{{ end }}`
```
jasonodonnell: One small request: add a newline to the end of the `.Data.id` line (like the default template) because if you were to look at it in the container:

```
/vault/secrets $ ls
ca.cert      db-creds     server.cert  server.key   token
/vault/secrets $ cat token
s.gXuAjfknTq0dK1i5s5rsjxpI/vault/secrets $
```

agent-inject/agent/annotations.go
```diff
@@ -35,6 +35,10 @@ const (
 	// If not provided, a default generic template is used.
 	AnnotationAgentInjectTemplate = "vault.hashicorp.com/agent-inject-template"
 
+	// AnnotationAgentInjectToken is the annotation key for injecting the token
+	// from auth/token/lookup-self
+	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token-only"
```
jasonodonnell: I wonder if `vault.hashicorp.com/agent-inject-token-only` is confusing: it suggests that we're _only_ going to render a token.  Maybe this should be renamed to `agent-inject-share-token`?

agent-inject/agent/annotations_test.go
```diff
@@ -218,6 +219,66 @@ func TestSecretTemplateAnnotations(t *testing.T) {
...
+					Name:     "token",
+					Path:     "auth/token/lookup-self",
+					Template: `{{ with secret "auth/token/lookup-self" }}{{ .Data.id }}{{ end }}`,
+				},
+			},
+		},
+		{
+			"invalid inject-token-only",
+			map[string]string{
+				"vault.hashicorp.com/agent-inject-token": "true",
```
jasonodonnell: Nit: use the annotation variable.

agent-inject/agent/annotations_test.go
```diff
@@ -218,6 +219,66 @@ func TestSecretTemplateAnnotations(t *testing.T) {
...
+func TestTemplateShortcuts(t *testing.T) {
+	tests := []struct {
+		name            string
+		annotations     map[string]string
+		expectedSecrets map[string]Secret
+	}{
+		{
+			"valid inject-token-only",
+			map[string]string{
+				"vault.hashicorp.com/agent-inject-token-only": "true",
```
jasonodonnell: Nit: use the annotation variable.

agent-inject/agent/annotations.go
tvoran: How about just `agent-inject-token`?

jasonodonnell: Works for me.

agent-inject/agent/config.go
tvoran: Added in d9dcbcddb2d915921112e0e98258fec3052bce40

agent-inject/agent/annotations.go
tvoran: Added in d9dcbcddb2d915921112e0e98258fec3052bce40

agent-inject/agent/annotations_test.go
tvoran: Changed in d9dcbcddb2d915921112e0e98258fec3052bce40

agent-inject/agent/config.go
tvoran: Hmm, it's literally writing the `\n` characters, not a new line.

tvoran: Fixed in c2a7199e572fa02d3ea60c1154b188f5842cb3e6

dcatalano-figure:
this might be a dumb question, my apologies. is the `lookup_self` token the same as the client token that lives in `/home/vault/.token` on the vault-agent and vault-agent-init containers. if so is there a reason why there wouldn't want to be a shared k8s volume and mount to that path on all three containers that could share the token among all pods in the container? say, if the above annotation is used? please excuse my naivety.

PR #79, opened 2020-02-18
Add projected ServiceAccount token support

dippynark:


jasonodonnell:
Hey @dippynark, thanks for the contribution!

I think this feature could be useful.  As it stands, the approach seems fine to me.  If you want to get this ready for review (cleanup as you mentioned, test coverage), we would gladly consider it for merge!

dippynark:
@jasonodonnell hey, sorry for the late reply, will start working on this soon

dippynark:
@jasonodonnell again. sorry for the slow again, but should be implemented now - the extra tests a quite simple so shout if you'd need to see more.

With this PR, vault-agent can use a projected service account token which be rotated by Kubernetes, potentially invalidating credentials and secrets retrieved using the old token. I assume vault-agent studies token TTLs and secret leases to ensure the rendered secrets are always valid and additionally will reload the rotating token as necessary, but I am not familiar with vault-agent code, so if this is not the case potentially some more work is needed.

dippynark:
@jasonodonnell is there anything else needed from this PR to get it merged? Or any reason why this functionality isn't needed/wanted anymore?

agent-inject/agent/agent.go
```diff
@@ -20,6 +20,7 @@ const (
 	DefaultAgentRunAsGroup         = 1000
 	DefaultAgentRunAsSameUser      = false
 	DefaultAgentSetSecurityContext = true
+	defaultServiceAccountPath      = "/var/run/secrets/vault.hashicorp.com/serviceaccount"
```
jasonodonnell, line 23: This should be `DefaultServiceAccountPath` to match style of other defaults.

Additionally, is this the correct value?  Typically this is `/var/run/secrets/kubernetes.io/serviceaccount` and I'm not understanding where `vault.hashicorp.com` is coming from.

agent-inject/agent/annotations.go
```diff
@@ -111,6 +111,15 @@ const (
...
 
+	// AnnotationAgentServiceAccountTokenVolumeName is the optional name of a volume containing a
+	// service account token
+	AnnotationAgentServiceAccountTokenVolumeName = "vault.hashicorp.com/agent-service-account-token-volume-name"
+
+	// AnnotationAgentServiceAccountTokenVolumePath is the optional path to a service account token
+	// within a volume named by "vault.hashicorp.com/agent-service-account-token-volume-name" or
+	// discovered by searching for service account volume mounts - the default is "token"
+	AnnotationAgentServiceAccountTokenVolumePath = "vault.hashicorp.com/agent-service-account-token-volume-path"
+
```
jasonodonnell, line 122: Anyway we can make these annotations shorter?  They're very close to max length.

agent-inject/agent/agent.go
dippynark: that const isn't used outside the package though so shouldn't it stay private? I think that applies to the other variables too, but just left them as they are

dippynark: the `defaultServiceAccountPath` would be the default for using projected service account tokens, as the ' normal' service account token is already being searched for - we can change this, but I think it'd make sense to have one that isn't the same as the normal path

agent-inject/agent/annotations.go
dippynark: @jasonodonnell any update on this? personally I don't see why we want to shorten the names as they are descriptive for a reason and they do fit within the limits, but if you have any other suggestions for names I am happy to change them

agent-inject/agent/agent.go
jasonodonnell: @dippynark Yes these aren't used outside of the package, but it should still follow style present with the other vars.

jasonodonnell: I missed the fallback search, so this should be OK.

agent-inject/agent/container_init_sidecar.go
```diff
@@ -19,8 +20,10 @@ func (a *Agent) ContainerInitSidecar() (corev1.Container, error) {
 			ReadOnly:  false,
 		},
 		{
-			Name:      a.ServiceAccountName,
-			MountPath: a.ServiceAccountPath,
+			Name: a.ServiceAccountTokenVolume.Name,
+			// Only mount the token
+			MountPath: path.Join(a.ServiceAccountTokenVolume.MountPath, a.ServiceAccountTokenVolume.TokenPath),
+			SubPath:   a.ServiceAccountTokenVolume.TokenPath,
```
jasonodonnell, line 26: When I manually tested this for backwards compatibility errors appeared in the injected containers:

```
2020-07-09T17:25:17.904Z [ERROR] auth.handler: error getting path or data from method: error="error reading JWT with Kubernetes Auth: read /var/run/secrets/kubernetes.io/serviceaccount/token: is a directory" backoff=1.178357559
```

dippynark: I guess a lot has changed since I wrote this - will have a look

dippynark:
@jasonodonnell should be good to go now - I just reversed my change to only mount the token

venky999:
any update on when this will be merged

dippynark:
@jasonodonnell any chance we could get this merged? Have rebased again

agent-inject/agent/agent.go
```diff
@@ -577,15 +590,34 @@ func (a *Agent) Validate() error {
 }
 
 func serviceaccount(pod *corev1.Pod) (string, string) {
-	var serviceAccountName, serviceAccountPath string
+
```
tvoran, line 593: Should this just return a ServiceAccountTokenVolume type?

I'd also like to see some unit tests for this function since the logic is getting a little complex.

agent-inject/agent/agent_test.go
```diff
@@ -66,19 +66,25 @@ func TestValidate(t *testing.T) {
 	}{
 		{
 			Agent{
-				Namespace:          "test",
-				ServiceAccountPath: "foobar",
-				ServiceAccountName: "foobar",
-				ImageName:          "test",
-				ConfigMapName:      "test",
+				Namespace: "test",
+				ServiceAccountTokenVolume: ServiceAccountTokenVolume{
```
tvoran, line 70: It looks like these all need to be type `*ServiceAccountTokenVolume`.

```suggestion
				ServiceAccountTokenVolume: &ServiceAccountTokenVolume{
```

tvoran (commented):
I like that this can mount the projected token directly on the vault-agent containers without copying from the app container! Left some suggestions here.

QuduSEC:
I'll just let you know that this feature is very relevant. And I also look forward to its implementation!

Issue #88, opened 2020-02-28
Metadata annotations are limited to 63 char, filename is embedded in annotation


sbeaulie:
I tried creating a filename that is longer than `foo` and I'm getting this error returned when creating the pods:
```metadata.annotations: Invalid value: "vault.hashicorp.com/agent-inject-secret-.aws_us_east_aws_secret_key.pem": name part must be no more than 63 characters```

While I understand this to be a k8s limitation (probably), the fact that the hashicorp prefix is 40 char long for `vault.hashicorp.com/agent-inject-secret-` and 42 char long for `vault.hashicorp.com/agent-inject-template-` it leaves very little space for the filename.

It also feels like bad practice to have the filename embedded in an annotation name.

**Proposed Solution**: 
Annotation values seem to accept arbitrarily long data sequence (I don't know the limit but my template definitions are long). 
Do not embed the filename in the annotation name, but rather in the value. 
Add another annotation for the expected filename and use that to create the file on disk. 

Example configuration after the change:
```
vault.hashicorp.com/agent-inject-secret-foo: "vault/secret/stored" # foo here is a unique identifier, not the filename
vault.hashicorp.com/agent-inject-secret-filename-foo: "my/aws/region/is/very/long/and/someone/loaded/the/pem/uuid/aaa-fffff-ggggg-eeee-awd-ffdgfdg-gfdgfdgfdg.pem" # stored in the value
vault.hashicorp.com/agent-inject-template-foo: "" # no change here
```

That way it has the extra advantage that the filename can contain a directory "/" delimiter since the annotation name can't.

I understand this is not a fun issue report to get as the solution above would be a breaking change for current users. As a compromise you could keep the current limited behavior, and intead use a new annotation for this split location / filename scenario:

```
vault.hashicorp.com/agent-inject-location-foo: "vault/secret/stored" # foo here is a unique identifier, not the filename. the value is the location of the vault secret
vault.hashicorp.com/agent-inject-filename-foo: "my/aws/region/is/very/long/and/someone/loaded/the/pem/uuid/aaa-fffff-ggggg-eeee-awd-ffdgfdg-gfdgfdgfdg.pem" # stored in the value
vault.hashicorp.com/agent-inject-template-foo: "" # no change here
```

the old (deprecated) way `vault.hashicorp.com/agent-inject-secret-<my_filename>` used by itself would still work but `<my_filename>` would be limited to 21 char if there is a template and 23 if there isn't one.

[...]

PR #89, opened 2020-02-28
Use separate annotation to specify the secret filename

sbeaulie:
as per issue https://github.com/hashicorp/vault-k8s/issues/88 before this fix the secret filename was limited to 21 chars and could not contain a directory separator "/"

This fix introduces two new annotations to use together

- `vault.hashicorp.com/agent-inject-location` to specify the vault secret location
- `vault.hashicorp.com/agent-inject-filename` to specify the filename to realize

It has the advantage of allowing long filenames, include directory separators
and if the location is absolute, it will create the file there, effectively giving
the option to save outside of /vault/secrets/

Note: backwards compatibility is respected for `vault.hashicorp.com/agent-inject-secret` knowing that it has limitations:
- filename length of 21 (or 23 if not using templates)
- cannot use "/" directory separator
- filename is forced to lowercase version, I do not know the original reason for this, but kept it here

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -20,18 +20,41 @@ const (
 	// be set to a true or false value, as parseable by strconv.ParseBool
 	AnnotationAgentInject = "vault.hashicorp.com/agent-inject"
 
-	// AnnotationAgentInjectSecret is the key annotation that configures Vault
+	// (Deprecated) AnnotationAgentInjectSecret is the key annotation that configures Vault
 	// Agent to retrieve the secrets from Vault required by the app.  The name
 	// of the secret is any unique string after "vault.hashicorp.com/agent-inject-secret-",
 	// such as "vault.hashicorp.com/agent-inject-secret-foobar".  The value is the
 	// path in Vault where the secret is located.
+	// Deprecated because an annotation is max 63 chars in kubernetes and also does not
+	// support directory separators "/"
+	// Any attempt to create a longer name or "/" with this annotation will result in the
+	// secret not being pulled, effectively skipping this annotation
+	// This annotation will always store the file in /vaults/secrets/
+	// This annotation force lowercase the selected name eg vault.hashicorp.com/agent-inject-secret-FOOBAR
+	// will realize the secret file in /vaults/secrets/foobar
 	AnnotationAgentInjectSecret = "vault.hashicorp.com/agent-inject-secret"
+
+	// AnnotationAgentInjectLocation is the annotation that contains the valid vault path to query the
+	// secret from. Anything after the final "-" is used as a unique id to match annotation rules together
+	// use in combination with AnnotationAgentInjectFilename
+	// eg. vault.hashicorp.com/agent-inject-location-foobar: "vault/secret/place"
+	//     vault.hashicorp.com/agent-inject-filename-foobar: "directory/containing/the_secret_filename.txt"
+	AnnotationAgentInjectLocation = "vault.hashicorp.com/agent-inject-location"
+
+	// AnnotationAgentInjectFilename is the annotation that contains the filename to create on disk
+	// Anything after the final "-" is used as a unique id to match annotation rules together
+	// use in combination with AnnotationAgentInjectLocation
+	// If you specify a filename with an absolute location like "/tmp/oreo/my_secret.pem"
+	// it will get created there, if not it defaults to using a root directory of "/vault/secrets/""
+	// eg. vault.hashicorp.com/agent-inject-location-foobar: "vault/secret/place"
+	//     vault.hashicorp.com/agent-inject-filename-foobar: "directory/containing/the_secret_filename.txt"
+	AnnotationAgentInjectFilename = "vault.hashicorp.com/agent-inject-filename"
 
 	// AnnotationAgentInjectTemplate is the key annotation that configures Vault
 	// Agent what template to use for rendering the secrets.  The name
 	// of the template is any unique string after "vault.hashicorp.com/agent-inject-template-",
 	// such as "vault.hashicorp.com/agent-inject-template-foobar".  This should map
-	// to the same unique value provided in ""vault.hashicorp.com/agent-inject-secret-".
+	// to the same unique value provided in "vault.hashicorp.com/agent-inject-location-".
 	// If not provided, a default generic template is used.
 	AnnotationAgentInjectTemplate = "vault.hashicorp.com/agent-inject-template"
 
@@ -42,7 +65,7 @@ const (
 	// AnnotationAgentInjectCommand is the key annotation that configures Vault Agent
 	// to run a command after the secret is rendered. The name of the template is any
 	// unique string after "vault.hashicorp.com/agent-inject-command-". This should map
-	// to the same unique value provided in ""vault.hashicorp.com/agent-inject-secret-".
+	// to the same unique value provided in "vault.hashicorp.com/agent-inject-location-".
 	// If not provided (the default), no command is executed.
 	AnnotationAgentInjectCommand = "vault.hashicorp.com/agent-inject-command"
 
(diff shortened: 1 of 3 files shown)


sbeaulie:
Of course I push this then realize there is a similar PR https://github.com/hashicorp/vault-k8s/pull/71 made by @smitthakkar96
My PR includes a way to define the "mountpath" by setting an absolute location (read starting with "/") and also allows for a long filenames with "/" in them. It preserves case sensitivity because it reads the value of the annotation as-is and does not apply any forced lowercase to it, so hopefully it also meets this requirement.

agent-inject/agent/annotations.go
```diff
@@ -20,18 +20,41 @@ const (
...
+	// use in combination with AnnotationAgentInjectFilename
+	// eg. vault.hashicorp.com/agent-inject-location-foobar: "vault/secret/place"
+	//     vault.hashicorp.com/agent-inject-filename-foobar: "directory/containing/the_secret_filename.txt"
+	AnnotationAgentInjectLocation = "vault.hashicorp.com/agent-inject-location"
+
+	// AnnotationAgentInjectFilename is the annotation that contains the filename to create on disk
+	// Anything after the final "-" is used as a unique id to match annotation rules together
+	// use in combination with AnnotationAgentInjectLocation
+	// If you specify a filename with an absolute location like "/tmp/oreo/my_secret.pem"
+	// it will get created there, if not it defaults to using a root directory of "/vault/secrets/""
```
smitthakkar96: `/vault/secrets` is a volume mount, if we are gonna do this we need to create additional volume mounts otherwise the secret won't be available in the main container

agent-inject/agent/annotations.go
```diff
@@ -20,18 +20,41 @@ const (
 	// be set to a true or false value, as parseable by strconv.ParseBool
 	AnnotationAgentInject = "vault.hashicorp.com/agent-inject"
 
-	// AnnotationAgentInjectSecret is the key annotation that configures Vault
+	// (Deprecated) AnnotationAgentInjectSecret is the key annotation that configures Vault
```
smitthakkar96: Since we are deprecating this, why not store json as the value of annotation, that way we can have more structured data when we process and also it would require less number of annotations. 

Example

``` yaml
vault.hashicorp.com/agent-inject-secret-config-FOOBAR: '{"secretPath": "secrets/service/database", "fileName": "database", "mountPath": "/etc/container_environment", "template": "<template>", "postInjectionCommand": "<optional command>"}'

PS: we can come up with better names 
```

agent-inject/agent/annotations.go
```diff
@@ -257,6 +284,39 @@ func secrets(annotations map[string]string) []*Secret {
...
 }
 
+func getNameFromAnnotation(name string, annotations map[string]string) string {
+	secretNamePrefix := fmt.Sprintf("%s-", AnnotationAgentInjectSecret)
+	locationSecretPrefix := fmt.Sprintf("%s-", AnnotationAgentInjectLocation)
+	if strings.Contains(name, secretNamePrefix) {
+		raw := strings.ReplaceAll(name, secretNamePrefix, "")
+		name = strings.ToLower(raw) // why would we force-lowercase the filename here?
+		if len(name) >= 21 || strings.Contains(name, "/") {
+			//houston we have a problem, skip secret, can we log this somewhere?
```
smitthakkar96: 👍

smitthakkar96: We should not skip but return a failure response IMHO. Errors should never pass silently. Container coming up with missing env var can fail or might run on a default that might cause a security issue.

agent-inject/agent/annotations.go
sbeaulie: Yes, of course

agent-inject/agent/annotations.go
sbeaulie: I can easily make that change but would like some guidance from the doce owners here. If we are to move to one annotation with metadata, should we still also support the old annotations? (I think not)
If we want to support the legacy annotations, and both are present, which one should take precedence?

agent-inject/agent/annotations.go
sbeaulie: Yes we can use the error class here.

agent-inject/agent/annotations.go
smitthakkar96: IMO new one should take precedence and we should support old one for backwards comparability and print warning in logs everytime it’s used

smitthakkar96: Throwing error incase both are present can also be an option. There is something definitely wrong if both are present.

agent-inject/agent/config.go
```diff
@@ -81,9 +82,16 @@ func (a *Agent) newTemplateConfigs() []*Template {
...
+
+		if strings.HasPrefix(secret.Name, "/") {
+			// it is only possible to use an absolute directory location with AnnotationAgentInjectLocation
+			destination = secret.Name
+		}
+
 		tmpl := &Template{
 			Contents:    template,
-			Destination: fmt.Sprintf("/vault/secrets/%s", secret.Name),
+			Destination: destination,
```
sbeaulie: based on further testing, it seems that there is a bug when the destination name contains a "/". Can someone help me trace this issue? Would it be in the vault agent code or the consul template code?

The issue is that the template does not seem to be applied if I chose a destination with that forward slash, we get the default 
```
{{ with secret \"%s\" }}{{ range $k, $v := .Data }}{{ $k }}: {{ $v }}\n{{ end }}{{ end }}
```
ie the file is created with the right filename containing the "/", but the template is not applied to it.

When I select a destination without "/" the template is applied properly.

sbeaulie: Also any filename longer than about 20 char gets created, but the same issue where the template does not get applied happens.

sbeaulie: @tvoran or @jasonodonnell if you could help me trace that issue I could fix it too! I took a look at the vault agent code and cannot find where that length limitation could be. The unit tests include filename lengths that are quite long too so I'm wondering if the issue is somewhere in the integration of vault-k8s to vault-agent?

smitthakkar96:
``` json
    {
      "destination": "/Users/smit.thakkar/projects/container_environment/CHAT_DJANGO_SECRET_KEY",
      "contents": "{{- with secret \"secret/beta/chat-service/chat-service-beta-tech/CHAT_DJANGO_SECRET_KEY\" -}} {{ .Data.value }} TEST_SMIT {{- end }}",
      "left_delimiter": "{{",
      "right_delimiter": "}}"
    },
```

I tried above and it worked for me.

smitthakkar96: I don't think filename longer than 20 char has any issue, what version are you using

smitthakkar96: when the destination name contains a "/" I get below error 
2020-03-06T16:49:22.480+0400 [ERROR] template.server: template server error: error="error rendering "(dynamic)" => "/Users/smit.thakkar/projects/container_environment/CHAT_DJANGO_SECRET_KEY/": failed reading file: open /Users/smit.thakkar/projects/container_environment/CHAT_DJANGO_SECRET_KEY/: not a directory"

OR

2020-03-06T16:49:09.889+0400 [ERROR] template.server: template server error: error="error rendering "(dynamic)" => "/Users/smit.thakkar/projects/container_environment/": failed reading file: read /Users/smit.thakkar/projects/container_environment/: is a directory"

tvoran:
Hi @sbeaulie, thanks for the work on this! 

I would suggest removing the `vault.hashicorp.com/agent-inject-location-` annotation, since the secret location in vault is already specified using the `vault.hashicorp.com/agent-inject-secret-<secret-name>` annotation. Then if the `vault.hashicorp.com/agent-inject-filename-` annotation is present, it would be used for the file name (and possibly path) when rendering the secret; otherwise the `<secret-name>` would be used for the filename under `/vault/secrets/`. I think this would probably involve adding a FileName field to the Secret struct. Does that make sense?

Let me know if you'd like to continue working on this, otherwise I can continue it in another branch.

tvoran:
Continued in #158

PR #91, opened 2020-03-02
Add annotation to change init container order

### LINKED ISSUE #53: Define order of vault init container
I have an application which uses an init container to do database migrations. This init container wants to have secrets from vault in it.

The vault init container is injected, but it never starts because the db migration sidecar starts first.

Is it possible to get the vault init container to run first?

jasonodonnell:
This adds a new annotation `vault.hashicorp.com/agent-init-first` which reorders the init container array to make the Vault Agent first.  Since init containers run sequentially in order they're defined, this makes when the Vault Agent runs configurable depending on different use cases (such as init containers that need vault secrets).

Also added secret mounts to other init containers present in the pod so they can consume Vault secrets if Vault Agent runs first.

Resolves #53 .

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -318,6 +318,8 @@ func (a *Agent) Patch() ([]byte, error) {
 			return patches, err
 		}
 
+		containers := a.Pod.Spec.InitContainers
+
 		// Init Containers run sequentially in Kubernetes and sometimes the order in
 		// which they run matters.  This reorders the init containers to put the agent first.
 		if a.InitFirst {
@@ -325,7 +327,7 @@ func (a *Agent) Patch() ([]byte, error) {
 			// Remove all init containers from the document so we can re-add them after the agent.
 			a.Patches = append(a.Patches, removeContainers("/spec/initContainers")...)
 
-			containers := []corev1.Container{container}
+			containers = []corev1.Container{container}
 			containers = append(containers, a.Pod.Spec.InitContainers...)
 
 			a.Patches = append(a.Patches, addContainers(
@@ -337,6 +339,17 @@ func (a *Agent) Patch() ([]byte, error) {
 				a.Pod.Spec.InitContainers,
 				[]corev1.Container{container},
 				"/spec/initContainers")...)
+		}
+
+		//Add Volume Mounts
+		for i, container := range containers {
+			if container.Name == "vault-agent-init" {
+				continue
+			}
+			a.Patches = append(a.Patches, addVolumeMounts(
+				container.VolumeMounts,
+				a.ContainerVolumeMounts(),
+				fmt.Sprintf("/spec/initContainers/%d/volumeMounts", i))...)
 		}
 	}
 
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -33,6 +33,9 @@ type Agent struct {
 	// in a pod request.
 	Inject bool
 
+	// InitFirst controls whether an init container is first to run.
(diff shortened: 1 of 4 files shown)


agent-inject/agent/agent.go
```diff
@@ -309,10 +317,27 @@ func (a *Agent) Patch() ([]byte, error) {
...
-		a.Patches = append(a.Patches, addContainers(
-			a.Pod.Spec.InitContainers,
-			[]corev1.Container{container},
-			"/spec/initContainers")...)
+
+		// Init Containers run sequentially in Kubernetes and sometimes the order in
+		// which they run matters.  This reorders the init containers to put the agent first.
+		if a.InitFirst {
+
+			// Remove all init containers from the document so we can readd them after the agent.
```
malnick: re-add

agent-inject/agent/agent.go
```diff
@@ -309,10 +317,27 @@ func (a *Agent) Patch() ([]byte, error) {
 		if err != nil {
 			return patches, err
 		}
-		a.Patches = append(a.Patches, addContainers(
-			a.Pod.Spec.InitContainers,
-			[]corev1.Container{container},
-			"/spec/initContainers")...)
+
+		// Init Containers run sequentially in Kubernetes and sometimes the order in
+		// which they run matters.  This reorders the init containers to put the agent first.
```
malnick: We may want to provide a brief sentence discussing the context for which a user would want to enable this.

agent-inject/agent/agent.go
jasonodonnell: Thanks!

agent-inject/agent/annotations.go
```diff
@@ -53,6 +53,10 @@ const (
 	// originated from.
 	AnnotationAgentRequestNamespace = "vault.hashicorp.com/agent-request-namespace"
 
+	// AnnotationAgentInitFirst makes the initialization container the first container
+	// to run when a pod starts.  Default is last.
```
malnick: Extra space between two sentences should be removed.

jasonodonnell:
Since it was relevant I also fixed a bug where the init containers weren't getting secret mount volumes.  Init containers can now consume Vault secrets if the Vault Agent runs first.

agent-inject/agent/agent.go
jasonodonnell: Done!

agent-inject/agent/annotations.go
jasonodonnell: Done!

agent-inject/agent/agent.go
```diff
@@ -309,10 +317,44 @@ func (a *Agent) Patch() ([]byte, error) {
...
+				a.Patches = append(a.Patches, removeContainers("/spec/initContainers")...)
+			}
+
+			containers = []corev1.Container{container}
+			containers = append(containers, a.Pod.Spec.InitContainers...)
+
+			a.Patches = append(a.Patches, addContainers(
+				[]corev1.Container{},
+				containers,
+				"/spec/initContainers")...)
```
tvoran: I'm curious why `containers` is created with all the containers before this, instead of doing something like:

```suggestion
			a.Patches = append(a.Patches, addContainers(
				[]corev1.Container{container},
				a.Pod.Spec.InitContainers,
				"/spec/initContainers")...)
```

jasonodonnell: This doesn't work due to the way `addContainers` is implemented.  The first arg is just the target, IE how many are existing currently.  Since we removed all containers before, we send an empty slice.

PR #97, opened 2020-03-03
added approle auto-auth functionality

rollerd:
- Adds the ability to use approle auto-auth method in the Vault agent rather than kubernetes
- Approle credentials are stored as kube secrets
- Adds the following annotations:
```
vault.hashicorp.com/agent-auto-auth-method: "approle"
vault.hashicorp.com/agent-approle-secret-name: "<approle_secret_name>"
```

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -93,6 +93,12 @@ type Agent struct {
 
 	// Vault is the structure holding all the Vault specific configurations.
 	Vault Vault
+
+	// AutoAuthMethod specifies whether to use default 'kubernetes' or 'approle' Vault Agent auto-auth methods
+	AutoAuthMethod string
+
+	// The name of the Kubernetes secret to lookup that contains the roleid and secretid for the approle
+	ApproleSecretName string
 }
 
 type Secret struct {
@@ -182,6 +188,8 @@ func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, erro
 		ServiceAccountName: saName,
 		ServiceAccountPath: saPath,
 		Status:             pod.Annotations[AnnotationAgentStatus],
+		AutoAuthMethod:     pod.Annotations[AnnotationAgentAutoAuthMethod],
+		ApproleSecretName:  pod.Annotations[AnnotationApproleSecretName],
 		Vault: Vault{
 			Address:          pod.Annotations[AnnotationVaultService],
 			AuthPath:         pod.Annotations[AnnotationVaultAuthPath],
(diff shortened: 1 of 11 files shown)


agent-inject/agent/agent.go
```diff
@@ -182,6 +188,8 @@ func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, erro
 		ServiceAccountName: saName,
 		ServiceAccountPath: saPath,
 		Status:             pod.Annotations[AnnotationAgentStatus],
+		AutoAuthMethod:     pod.Annotations[AnnotationAgentAutoAuthMethod],
```
smitthakkar96: It's a great idea to add support for other auto auth methods. IMO it's better to have AutoAuthConfig which takes json and you can set the params to not have large number of annotations. Thoughts?

agent-inject/agent/agent.go
```diff
@@ -272,6 +280,14 @@ func ShouldInject(pod *corev1.Pod) (bool, error) {
 func (a *Agent) Patch() ([]byte, error) {
 	var patches []byte
 
+	// Add our vault-approle-secrets volume
+	if a.AutoAuthMethod == "approle" {
```
smitthakkar96: Would it be better to abstract this and have an interface that has two methods `GetAutoAuthPatches` and `GetAutoAuthConfig`

agent-inject/agent/config.go
```diff
@@ -94,6 +94,18 @@ func (a *Agent) newTemplateConfigs() []*Template {
 }
 
 func (a *Agent) newConfig(init bool) ([]byte, error) {
+	config_map := make(map[string]interface{})
+	switch a.AutoAuthMethod {
+	case "approle":
+		config_map["role_id_file_path"] = "/vault/approle/roleid"
+		config_map["secret_id_file_path"] = "/vault/approle/secretid"
+		config_map["remove_secret_id_file_after_reading"] = false
+	case "kubernetes":
```
smitthakkar96: This could be replaced by the `GetAutoAuthConfig`

Have two structs implement this interface `KubernetesAutoAuth` and `AppRoleAutoAuth`.

rollerd:
@smitthakkar96  - thanks very much for the feedback! I'm hoping to have some time this week to dig and make the suggested changes.

rollerd:
Closing per the conversation and better options discussed in issue #115

adawalli:
Really appreciate the work you did on https://github.com/hashicorp/vault-k8s/pull/119 @rollerd - I still wish that we could have also had the option for annotations. The only change I am making is enabling approle, but that means we can't use any of the other fancy annotations for grabbing secrets, etc. Your patch allows approle to work, which is excellent...I just wish for a cleaner deployment (I personally like this original PR better :P )

PR #112, opened 2020-03-31
Using MutatingWebhook v1 api

eye0fra:
I upgraded MutatingWebhookConfiguration v1 API. 
AdmissionReview response is compatible with Kubernetes 1.16.2 and OpenShift 4.3.x.
This change is to avoid an undefined AdmissionReview response after the injection.
The k8s API library needs to be updated.


smitthakkar96:
@eye0fra will this change work with k8s 1.15

eye0fra:
Hi @smitthakkar96,
The v1 version has been introduced on k8s 1.16.
I can just rollback the MutatingWebhookConfiguration manifest deploy/injector-mutating-webhook.yaml  so it would work as well with 1.15.

smitthakkar96:
@eye0fra @jasonodonnell IMO we should ensure backwards compatibility for 1.15 as Amazon EKS only supports upto kubernetes 1.15 and lot of people use EKS and Vault Injector including dubizzle. That being said I would suggest either rollback the MutatingWebhookConfiguration manifest or update vault injector to handle both versions for time being.

deploy/injector-mutating-webhook.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
```
smitthakkar96, line 2: what about people who install via the helm chart

eye0fra: via helm chart is v1beta1.
https://github.com/hashicorp/vault-helm/blob/master/templates/injector-mutating-webhook.yaml
It can be enhanced to use both.

eye0fra:
I removed the MutatingWebhookConfiguration manifest.

elsesiy:
I tested this change but the injector fails to inject the sidecar after. The agent injector currently doesn't implement the v1 api directly, see https://github.com/hashicorp/vault-k8s/blob/a9ff2f2a7fa66855166415248561c77e0662a910/agent-inject/handler.go#L81

swenson:
Thanks for this. I think this is no longer necessary as we have upgraded elsewhere to the `admissionregistration.k8s.io/v1` and are compatible with either in `injector.go`.

PR #119, opened 2020-04-09
added extra-secret annotation for adding kube-secrets

rollerd:
- Adds annotation:
 ```vault.hashicorp.com/agent-extra-secret: <somekubesecret>```
which will mount the supplied Kubernetes secret as a volume at `/vault/custom` in the sidecar/init containers. Useful for custom Agent configs with auto-auth methods such as approle that require paths to secrets be present.

- To use: create a custom agent configmap (with config.hcl or config-init.hcl as secret keys), create a deployment with the new annotation, create a Kube secret with the roleid/secretid/other secret you reference from your configmap (/vault/custom/somesecret).


karras:
What is the status here? We also would prefer using Kube secrets instead of directly writing files.

tvoran (approved):
Thanks for the patch! And apologies for the delay.

lcgkm:
Awesome! I look forward to the official release of this feature.

adawalli:
same here @lcgkm - we have a Vault Endpoint that "in the cloud" but our k8s cluster is on prem and thus the k8s auth just doesn't work for us. 

I am also hoping we can get a **learn** article (or even just a Readme.md in this repo) for examples on how this feature can be used (specifically against something like approle!)

PR #131, opened 2020-05-05
New securityContext options and annotations

tvoran:
(Continuation of #126)

Adds an option for running the injected vault containers as the User of the first application container in the Pod. Requires `Spec.Containers[0].SecurityContext.RunAsUser` to be set in the pod spec. Defaults to false. Available as:
- `-run-as-same-user` command-line option
- `AGENT_INJECT_RUN_AS_SAME_USER` environment variable
- `vault.hashicorp.com/agent-run-as-same-user` annotation

Also adds an option for controlling whether or not the SecurityContext is set on injected containers. Defaults to true. Available as:
- `-set-security-context` command-line option
- `AGENT_INJECT_SET_SECURITY_CONTEXT` environment variable
- `vault.hashicorp.com/agent-set-security-context` annotation

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -166,6 +166,7 @@ type AgentConfig struct {
 	RevokeOnShutdown bool
 	UserID           string
 	GroupID          string
+	SameID           bool
 }
 
 // Init configures the expected annotations required to create a new instance
@@ -244,8 +245,12 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 	}
 
 	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentRunAsUser]; !ok {
+
 		if cfg.UserID == "" {
 			cfg.UserID = strconv.Itoa(DefaultAgentRunAsUser)
+		}
+		if cfg.SameID {
+			cfg.UserID = strconv.FormatInt(*pod.Spec.Containers[0].SecurityContext.RunAsUser, 10)
 		}
 		pod.ObjectMeta.Annotations[AnnotationAgentRunAsUser] = cfg.UserID
 	}
@@ -253,6 +258,9 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentRunAsGroup]; !ok {
 		if cfg.GroupID == "" {
 			cfg.GroupID = strconv.Itoa(DefaultAgentRunAsGroup)
+		}
+		if cfg.SameID {
+			cfg.GroupID = strconv.Itoa(0)
 		}
 		pod.ObjectMeta.Annotations[AnnotationAgentRunAsGroup] = cfg.GroupID
 	}
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -448,6 +448,9 @@ func (a *Agent) runAsSameID(pod *corev1.Pod) (bool, error) {
 		if pod.Spec.Containers[0].SecurityContext.RunAsUser == nil {
 			return DefaultAgentRunAsSameUser, errors.New("RunAsUser is nil for Container 0's SecurityContext")
 		}
+		if *pod.Spec.Containers[0].SecurityContext.RunAsUser == 0 {
+			return DefaultAgentRunAsSameUser, errors.New("container not allowed to run as root")
+		}
 		a.RunAsUser = *pod.Spec.Containers[0].SecurityContext.RunAsUser
 	}
 	return runAsSameID, nil
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -250,6 +250,15 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 			cfg.UserID = strconv.Itoa(DefaultAgentRunAsUser)
 		}
 		if cfg.SameID {
+			if len(pod.Spec.Containers) == 0 {
+				return errors.New("No containers found in Pod Spec")
+			}
+			if pod.Spec.Containers[0].SecurityContext == nil {
+				return errors.New("No SecurityContext found for Container 0")
+			}
+			if pod.Spec.Containers[0].SecurityContext.RunAsUser == nil {
+				return errors.New("RunAsUser is nil for Container 0's SecurityContext")
+			}
 			cfg.UserID = strconv.FormatInt(*pod.Spec.Containers[0].SecurityContext.RunAsUser, 10)
 		}
 		pod.ObjectMeta.Annotations[AnnotationAgentRunAsUser] = cfg.UserID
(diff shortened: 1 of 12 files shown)


tvoran:
Though I'm wondering if this should be an annotation in addition to (or instead of) the command-line option.

eye0fra:
To have the same behavior like the others command line I think is a good idea.

agent-inject/agent/annotations.go
```diff
@@ -244,9 +245,22 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
+			if len(pod.Spec.Containers) == 0 {
+				return errors.New("No containers found in Pod Spec")
+			}
+			if pod.Spec.Containers[0].SecurityContext == nil {
+				return errors.New("No SecurityContext found for Container 0")
+			}
+			if pod.Spec.Containers[0].SecurityContext.RunAsUser == nil {
+				return errors.New("RunAsUser is nil for Container 0's SecurityContext")
+			}
+			cfg.UserID = strconv.FormatInt(*pod.Spec.Containers[0].SecurityContext.RunAsUser, 10)
```
tvoran, line 262: I wonder if this should be looking at `pod.Spec.SecurityContext` as well.

eye0fra: The `Spec.SecurityContext` is common for all containers running inside the pod. 
The `Spec.Container[0].SecurityContext` overrides the settings at the pod level.

tvoran: Yep, and after running this on OpenShift via CRC, I see that `RunAsUser` is being set at the container level.

I'm also now thinking that it might be simpler to just not set a SecurityContext at all on the init and sidecar containers, since OpenShift will do that for you based on the project/namespace.

eye0fra: Mmm I have to double check it, but the reason that we explicitly put the security context is because we are injecting the side/init-container with the webhook and it is not OpenShift doing it.

tvoran:
> To have the same behavior like the others command line I think is a good idea.

Indeed, so now RunAsSameUser is available as both a command-line option and an annotation, to match the user and group options.

I also added a SetSecurityContext command-line option and annotation, to control whether the securityContext is set or not. Setting this to `false` also works in my OpenShift setup.

agent-inject/agent/annotations.go
```diff
@@ -101,6 +101,16 @@ const (
 	// AnnotationAgentRunAsGroup sets the Group ID to run the Vault Agent containers as.
 	AnnotationAgentRunAsGroup = "vault.hashicorp.com/agent-run-as-group"
 
+	// AnnotationAgentRunAsSameUser sets the User ID of the injected Vault Agent
+	// containers to the User ID of the first application container in the Pod.
+	// Requires Spec.Containers[0].SecurityContext.RunAsUser to be set in the
+	// Pod Spec.
+	AnnotationAgentRunAsSameUser = "vault.hashicorp.com/agent-run-as-same-user"
```
jasonodonnell, line 108: We're not guaranteed that `Container[0]` is the app container you want to mirror the user of.  I'm thinking pods that have a more than one sidecar container predefined.  This could be a documentation note, but worth thinking about.

agent-inject/agent/annotations.go
```diff
@@ -394,3 +428,36 @@ func (a *Agent) preserveSecretCase(secretName string) (bool, error) {
...
+		if len(pod.Spec.Containers) == 0 {
+			return DefaultAgentRunAsSameUser, errors.New("No containers found in Pod Spec")
+		}
+		if pod.Spec.Containers[0].SecurityContext == nil {
+			return DefaultAgentRunAsSameUser, errors.New("No SecurityContext found for Container 0")
+		}
+		if pod.Spec.Containers[0].SecurityContext.RunAsUser == nil {
+			return DefaultAgentRunAsSameUser, errors.New("RunAsUser is nil for Container 0's SecurityContext")
+		}
+		a.RunAsUser = *pod.Spec.Containers[0].SecurityContext.RunAsUser
```
jasonodonnell, line 451: We should reject values that make our container run as root.  Thoughts?

agent-inject/agent/annotations.go
```diff
@@ -394,3 +428,36 @@ func (a *Agent) preserveSecretCase(secretName string) (bool, error) {
...
+	raw, ok := a.Annotations[AnnotationAgentRunAsSameUser]
+	if !ok {
+		return DefaultAgentRunAsSameUser, nil
+	}
+	runAsSameID, err := strconv.ParseBool(raw)
+	if err != nil {
+		return DefaultAgentRunAsSameUser, err
+	}
+	if runAsSameID {
+		if len(pod.Spec.Containers) == 0 {
```
jasonodonnell, line 442: Probably OK to have it but can this ever be true?

agent-inject/agent/annotations_test.go
```diff
@@ -9,14 +9,16 @@ import (
...
 	"github.com/mattbaird/jsonpatch"
+	"github.com/stretchr/testify/require"
 )
 
 func TestInitCanSet(t *testing.T) {
 	annotations := make(map[string]string)
 	pod := testPod(annotations)
 
-	err := Init(pod, AgentConfig{"foobar-image", "http://foobar:8200", "test", "test", true, "1000", "100"})
+	err := Init(pod, AgentConfig{"foobar-image", "http://foobar:8200", "test", "test", true, "1000", "100", DefaultAgentRunAsSameUser, DefaultAgentSetSecurityContext})
```
jasonodonnell, line 21: It might be time to define this AgentConfig struct outside of the function call.  It's getting rather large.

agent-inject/agent/annotations.go
tvoran: Yeah, that's a good point. Perhaps this option could specify the name of the container to mirror, instead of just grabbing `Container[0]`. Though that makes it hard to have as a command-line option. 

Maybe a future annotation for specifying the name of the container to grab, if there's demand?

agent-inject/agent/annotations.go
tvoran: That's fine with me, and easy to change if there's somehow demand to run as root.

agent-inject/agent/annotations.go
tvoran: Yeah that's just to protect the next line, since it's referencing `[0]` directly.

agent-inject/agent/annotations_test.go
tvoran: Yep, done in 3c1cf60.

agent-inject/agent/annotations.go
tvoran: Added in 3c1cf60.

agent-inject/agent/annotations.go
jasonodonnell: Yeah let's just wait to see if folks want it.  I don't think it's unreasonable to suggest putting the container they want first in the spec for now.

Issue #73, opened 2020-02-07
Support running Vault Agent cache in sidecar container


lawliet89:
https://www.vaultproject.io/docs/agent/caching/index.html

hamishforbes:
It would be helpful to have an annotation that enables a `listener {}` and `cache {}` block in the sidecar agent.

Something like

```
vault.hashicorp.com/agent-listen: bool (default: false)
vault.hashicorp.com/agent-listen-tls-disable: bool (default: true)
vault.hashicorp.com/agent-listen-address: string (default: 127.0.01:8200)
vault.hashicorp.com/agent-cache-auto-auth: bool (default: false)
```

I'm migrating a number of services onto k8s which previously were able to access Vault via an agent listening on plain HTTP bound to 127.0.0.1.

I can reproduce this in k8s but only by using the configmap method to add a listener, which is ok but very verbose and repetitive for many services

lawliet89:
#132 implements this. But I have kept TLS disabled always for now because there is no way to mount arbitrary volumes now.

PR #132, opened 2020-05-06
Add annotations to configure Agent Cache

lawliet89:
- TLS is always disabled because there is no way to mount arbitrary
volumes to the agent sidecar container

Fixes https://github.com/hashicorp/vault-k8s/issues/73

```
$ make test
go test -race ./...
?       github.com/hashicorp/vault-k8s  [no test files]
ok      github.com/hashicorp/vault-k8s/agent-inject     (cached)
ok      github.com/hashicorp/vault-k8s/agent-inject/agent       (cached)
ok      github.com/hashicorp/vault-k8s/helper/cert      (cached)
ok      github.com/hashicorp/vault-k8s/subcommand/injector      (cached)
?       github.com/hashicorp/vault-k8s/subcommand/version       [no test files]
?       github.com/hashicorp/vault-k8s/version  [no test files]
```

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -165,8 +165,8 @@ const (
 	// auto auth token or not. Can be set to "force" to force usage of the auto-auth token
 	AnnotationAgentCacheUseAutoAuthToken = "vault.hashicorp.com/agent-cache-use-auth-auth-token"
 
-	// AnnotationAgentCacheListenerAddress configures the address the agent cache should listen on
-	AnnotationAgentCacheListenerAddress = "vault.hashicorp.com/agent-cache-listener-address"
+	// AnnotationAgentCacheListenerPort configures the port the agent cache should listen on
+	AnnotationAgentCacheListenerPort = "vault.hashicorp.com/agent-cache-listener-port"
 )
 
 type AgentConfig struct {
@@ -272,8 +272,8 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationAgentCacheEnable] = DefaultAgentCacheEnable
 	}
 
-	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentCacheListenerAddress]; !ok {
-		pod.ObjectMeta.Annotations[AnnotationAgentCacheListenerAddress] = DefaultAgentCacheListenerAddress
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentCacheListenerPort]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationAgentCacheListenerPort] = DefaultAgentCacheListenerPort
 	}
 
 	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentCacheUseAutoAuthToken]; !ok {
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -156,6 +156,17 @@ const (
 	// AnnotationPreserveSecretCase if enabled will preserve the case of secret name
 	// by default the name is converted to lower case.
 	AnnotationPreserveSecretCase = "vault.hashicorp.com/preserve-secret-case"
+
+	// AnnotationAgentCacheEnable if enabled will configure the sidecar container
+	// to enable agent caching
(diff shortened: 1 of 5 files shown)


agent-inject/agent/annotations.go
```diff
@@ -156,6 +156,17 @@ const (
...
+	// AnnotationAgentCacheEnable if enabled will configure the sidecar container
+	// to enable agent caching
+	AnnotationAgentCacheEnable = "vault.hashicorp.com/agent-cache-enable"
+
+	// AnnotationAgentCacheUseAutoAuthToken configures the agent cache to use the
+	// auto auth token or not. Can be set to "force" to force usage of the auto-auth token
+	AnnotationAgentCacheUseAutoAuthToken = "vault.hashicorp.com/agent-cache-use-auth-auth-token"
+
+	// AnnotationAgentCacheListenerAddress configures the address the agent cache should listen on
+	AnnotationAgentCacheListenerAddress = "vault.hashicorp.com/agent-cache-listener-address"
```
jasonodonnell, line 169: Maybe this shouldn't be configurable but always be `127.0.0.1`.  This really shouldn't be listening to anything other than localhost.  Especially since the `use-auth-auth-token` parameter is configurable, unauthenticated clients could retrieve secrets from Vault if misconfigured.

lawliet89: I was thinking of allowing the user to change the port. Should I change this into a port configuration then?

jasonodonnell: What is your thoughts with enabling the listener?  If other containers in the pod want to use the vault agent as a proxy?

lawliet89: I am not sure what you mean by this question. Do you mean what are my thoughts on allowing the listener to listen to IP addresses other than `127.0.0.1`? I think you're right that it is a bad idea.

jasonodonnell: Just trying to understand the feature more in general.  We definitely don't want to bind to anything other than localhost, port configuration could be useful.

lawliet89: I see. I think this is a very useful feature. Other than just proxying, the vault agent cache also takes care of renewing secrets. I have some applications that basically "outsources" the renewal duty to Vault agent.

jasonodonnell: Agreed with it being useful.  I think turning this into a port configurable makes more sense.

lawliet89: OK. I will make the change in a bit.

lawliet89: Done. See https://github.com/hashicorp/vault-k8s/pull/132/commits/f0dc6868eb5e7f120496cfb43526f861139aee1f

lawliet89:
@jasonodonnell Could we get this merged in if it's ok? There's been quite a few conflicts.

hamishforbes:
Any chance of a release with this feature? It just barely missed the 0.4.0 release
Would really be appreciated as doing this with config maps for vault is a real pita!

reegnz:
Please do a 0.5.0 release with this! Would be really great if I could start using this and not have to do explicit agent configs to do this. :)

lawliet89:
@tvoran and @jasonodonnell : Could you please make a new release with this?

In any case, if you cannot wait, I've built and push a new image to `lawliet89/vault-k8s:0.4.1-gh132` from commit 23879d1fde10369dbeca559a82b547f5ebcf80ca

The caveats are:

- Built with `make image`
- Binary was compiled on my machine
- Binary is not signed by HashiCorp GPG key

jasonodonnell:
We're planning a release soon @lawliet89, there's a few more PRs we're hoping to pull in for that release.  Likely next week (August 17th-21st).

reegnz:
It seems like those images got pushed 3 days ago.
Hoping for a release today! 🤞

How come this release is blocked by the new image version BTW? These releases shouldn't rely on what vault version we're running.

jasonodonnell:
@reegnz, it doesn't really matter, but there are references internally for default injected image: https://github.com/hashicorp/vault-k8s/blob/1ee5b2e7916ec9b98dbf29679dccf864a76eb5c4/agent-inject/agent/agent.go#L17.

Whenever possible we try to update this to the latest image. 1.5.2 released late on Friday so Vault K8s releases today.

PR #142, opened 2020-06-03
Add readOnlyRootFilesystem to injected agents

jasonodonnell:
This adds `readOnlyRootFilesystem: true` to injected Agent containers for extra security.  We do some runtime configuration of the Vault Agent configuration file, so that needs to be done in a writable place.  Since a memory volume is mounted to `/home/vault`, I updated args to use it.


agent-inject/agent/container_sidecar.go
```diff
@@ -151,12 +151,14 @@ func (a *Agent) createLifecycle() corev1.Lifecycle {
...
 		runAsNonRoot = false
 	}
 	return &corev1.SecurityContext{
-		RunAsUser:    pointerutil.Int64Ptr(a.RunAsUser),
-		RunAsGroup:   pointerutil.Int64Ptr(a.RunAsGroup),
-		RunAsNonRoot: pointerutil.BoolPtr(runAsNonRoot),
+		RunAsUser:              pointerutil.Int64Ptr(a.RunAsUser),
+		RunAsGroup:             pointerutil.Int64Ptr(a.RunAsGroup),
+		RunAsNonRoot:           pointerutil.BoolPtr(runAsNonRoot),
+		ReadOnlyRootFilesystem: pointerutil.BoolPtr(true),
```
catsby, line 162: Should we use `DefaultAgentReadOnlyRoot` here instead of `true`?

agent-inject/agent/container_sidecar.go
```diff
@@ -15,7 +15,7 @@ const (
 	DefaultResourceLimitMem   = "128Mi"
 	DefaultResourceRequestCPU = "250m"
 	DefaultResourceRequestMem = "64Mi"
-	DefaultContainerArg       = "echo ${VAULT_CONFIG?} | base64 -d > /tmp/config.json && vault agent -config=/tmp/config.json"
+	DefaultContainerArg       = "echo ${VAULT_CONFIG?} | base64 -d > /home/vault/config.json && vault agent -config=/home/vault/config.json"
```
catsby, line 18: Does this have any implications if a user already has Vault setup with a prior version of the helm chart, and then they upgrade the helm chart? Will the `/home` dir exist?

jasonodonnell: It exists already and when they upgrade it will reinject the the containers with the new configuration.

agent-inject/agent/container_sidecar.go
jasonodonnell: Good catch, thought I got all of them.

Issue #17, opened 2019-12-20
Add support for exposing cert metrics (minted / expiry time etc.)


shank7485:
It would be great if various metrics can be exposed by like an endpoint `/metrics` for a scraper like Prometheus to scrap off. This is useful to monitor for cert minted and expiry times and use that to warn / alert on if necessary.

Here is an example project which is follows a similar pattern of using sidecars to mint and mount certificates from vault: https://github.com/monzo/vault-sidekick/ and the commit: https://github.com/monzo/vault-sidekick/commit/2e7365b733311a6b84e2e4bb6fadcd473815bfd8 exposes metrics in prometheus time series format. 

If similar metric is exposed in the vault-k8s, it would be great since this side-car is directly developed from Hashicorp / Vault and would fit in well with other Hashicorp infrastructure.

jasonodonnell:
@shank7485, this is a reasonable request.  We could probably make this happen in the new year and are open to PRs for this feature. 👍

PR #145, opened 2020-06-09
Export prometheus metrics

diogonicoleti:
I would like to monitor some behaviors of the injector, so I added the `metrics` endpoint that export some prometheus metrics.

Just for now, this just export the default metrics for go apps, but we can add injector specific metrics in the future (like this one: https://github.com/hashicorp/vault-k8s/issues/17).


diogonicoleti:
cc/ @jasonodonnell

diogonicoleti:
@jasonodonnell done! I added a new flag `telemetry-path` to set the metrics path. If this flag is not set, the prometheus handler is not registered, so we don't have any issue.

PR #146, opened 2020-06-10
Add annotation to change secret permission

### LINKED ISSUE #136: Lack of control of file permissions of rendered secrets
out of the box, vault injector renders secrets as:
uid: 100 gid: 1000 file mode: 0644
These are pretty ordinary defaults.

Consider that many applications that use lets say a cert key etc expect permissions of 0400
The uid/gid control is already handled in the unreleased commits for using:
```
        vault.hashicorp.com/agent-run-as-user: "1000"
        vault.hashicorp.com/agent-run-as-group: "1000"
```

The only work around i managed for permissions of the rendered files is the following:
```
        vault.hashicorp.com/agent-inject-command-${secret}: "chmod 400 ${secret_volume_path}/${secret}
```

Proposal:
Just with many of the other annotations:
1. set a default render file permission in the agent (ie not in annotations)
2. allow annotations to set a default file permission
3. allow per secret to set a file permission

Cheers!

diogonicoleti:
This adds a new annotation:
```yaml
vault.hashicorp.com/agent-inject-permission-
```
which changes the permission of the rendered secret.

Resolves #136

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -136,6 +136,9 @@ type Secret struct {
 	// Mount Path
 	MountPath string
 
+	// Permission is the optional permission to render the secret
+	Permission string
+
 	// Command is the optional command to run after rendering the secret.
 	Command string
 }
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -35,6 +35,14 @@ const (
 	// If not provided, a default generic template is used.
 	AnnotationAgentInjectTemplate = "vault.hashicorp.com/agent-inject-template"
 
+	// AnnotationAgentInjectPermission is the key annotation that configures Vault
+	// Agent what permission to render the secret.  The name
+	// of the template is any unique string after "vault.hashicorp.com/agent-inject-permission-",
+	// such as "vault.hashicorp.com/agent-inject-permission-foobar".  This should map
+	// to the same unique value provided in ""vault.hashicorp.com/agent-inject-secret-".
+	// If not provided, a default permission is used.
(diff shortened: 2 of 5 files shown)


jasonodonnell:
@diogonicoleti This looks great and is a good feature.  I'm wondering if we need to validate the permission bits before injection.  I think this could be done using https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/filemode#New.

Thoughts?

diogonicoleti:
@jasonodonnell I believe it's a good idea. I just was thinking what to do when we have an invalid permission bits. I see the following alternatives:

- Ignore the permission annotation, as if it is not set.
- Returns an error in `agent.go`: 
```go
agent.Secrets, err = agent.secrets()
if err != nil {
	return agent, err
}
```

To validate we can do something like this (that is similar how the `srd-d` filemode do):
```go
func validate(perm string) (string, error) {
	i, err := strconv.ParseInt(perm, 8, 32)
	if err != nil {
		return perm, err
	}
	return fmt.Sprintf("%04o", i), nil
}
```
It isn't a perfect validation because it just validate if the value is an octal number. I checked how `consul-template` do it, but I not found a validation for it. So I believe that this validation is better than nothing.

What do you think?

jasonodonnell:
I think the proposed validation would work here.

jasonodonnell:
Bumping this.

diogonicoleti:
@jasonodonnell Sorry for the delay I'm in a little busy now, but in the beginning of the next week I'm going to finish this PR! 😄

der-eismann:
@diogonicoleti Any updates on this? :slightly_smiling_face:

tvoran:
Superseded by #277

PR #147, opened 2020-06-10
Add golangci-lint

diogonicoleti:


jasonodonnell:
We have some upcoming CI work that will add similar checks, so closing.

PR #153, opened 2020-06-24
Add extra security contexts to injected containers

jasonodonnell:
Adding additional security contexts for better integration with restrictive PSP policies that require these to be set.

### DIFF
--- a/agent-inject/agent/container_sidecar.go
+++ b/agent-inject/agent/container_sidecar.go
@@ -160,5 +160,9 @@ func (a *Agent) securityContext() *corev1.SecurityContext {
 		RunAsGroup:             pointerutil.Int64Ptr(a.RunAsGroup),
 		RunAsNonRoot:           pointerutil.BoolPtr(runAsNonRoot),
 		ReadOnlyRootFilesystem: pointerutil.BoolPtr(DefaultAgentReadOnlyRoot),
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{DefaultAgentDropCapabilities},
+		},
+		AllowPrivilegeEscalation: pointerutil.BoolPtr(DefaultAgentAllowPrivilegeEscalation),
 	}
 }
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ BUILD_DIR=.build
 GOOS?=linux
 GOARCH?=amd64
 BIN_NAME=$(IMAGE_NAME)_$(GOOS)_$(GOARCH)_$(VERSION)
+GOFMT_FILES?=$$(find . -name '*.go' | grep -v vendor)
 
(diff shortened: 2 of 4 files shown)


agent-inject/agent/container_sidecar.go
```diff
@@ -160,5 +160,9 @@ func (a *Agent) securityContext() *corev1.SecurityContext {
 		RunAsGroup:             pointerutil.Int64Ptr(a.RunAsGroup),
 		RunAsNonRoot:           pointerutil.BoolPtr(runAsNonRoot),
 		ReadOnlyRootFilesystem: pointerutil.BoolPtr(DefaultAgentReadOnlyRoot),
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{DefaultAgentDropCapabilities},
```
tvoran, line 164: Will dropping all capabilities prevent actions like running a kill command in an agent-inject-command?

jasonodonnell: This is a really good question, but it appears not.  My app from HashiConf receives signals which the sidecar sends without issue when these capabilities are applied:

```
kubectl get pod app... -o yaml
  - name: vault-agent
     securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      readOnlyRootFilesystem: true
      runAsGroup: 1000
      runAsNonRoot: true
      runAsUser: 100
     ...
```

```
2020-06-26T16:01:20.046Z [INFO]  app: Listening on:: address=:8080
2020-06-26T16:01:20.387Z [INFO]  app: SIGHUP received. Reloading configuration..
2020-06-26T16:01:34.527Z [INFO]  app: SIGHUP received. Reloading configuration..
2020-06-26T16:01:48.437Z [INFO]  app: SIGHUP received. Reloading configuration..
2020-06-26T16:02:02.388Z [INFO]  app: SIGHUP received. Reloading configuration..
2020-06-26T16:02:16.379Z [INFO]  app: SIGHUP received. Reloading configuration..
```

tvoran: Cool, I did some playing with this too, and it looks like with or without cap_kill, the uid of the process issuing the kill needs to be the same as the target process. So I agree this doesn't change behavior with respect to kill:

```
## inside a vault-agent sidecar container, app running as 101, vault-agent as 100
/ $ ps auwwx
PID   USER     TIME  COMMAND
    1 root      0:00 /pause
   29 101       0:00 /app
   43 vault     0:00 vault agent -config=/home/vault/config.json
   74 vault     0:00 sh
   83 vault     0:00 ps auwwx
/ $ kill -HUP $(pidof app)
sh: can't kill pid 29: Operation not permitted
/ $ id
uid=100(vault) gid=1000(vault)
```

PR #158, opened 2020-07-23
Add file annotation

### LINKED ISSUE #88: Metadata annotations are limited to 63 char, filename is embedded in annotation

tvoran:
(Continues https://github.com/hashicorp/vault-k8s/pull/89, fixes #88)

Adds new annotation `vault.hashicorp.com/agent-inject-file`. This sets the filename and path in the secrets volume where a vault secret will be written. The container mount path for the secrets volume may be modified with the [`secret-volume-path`](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-secret-volume-path) annotation.

This allows for filenames that aren't limited in length as k8s annotation keys, and for arbitrary directory structures in the secrets volume.


agent-inject/agent/config.go
```diff
@@ -95,9 +96,16 @@ func (a *Agent) newTemplateConfigs() []*Template {
 			template = fmt.Sprintf(DefaultTemplate, secret.Path)
 		}
 
+		filePathAndName := secret.FilePathAndName
+		if filePathAndName == "" {
+			filePathAndName = fmt.Sprintf("%s/%s", secret.MountPath, secret.Name)
+		} else {
+			filePathAndName = filepath.Join(secret.MountPath, filePathAndName)
+		}
```
catsby, line 104: I think `filePathAndName := secret.FilePathAndName` ends up being ineffective because in either condition it immediately gets overwritten. I suggest refactoring to something like so:

```go
filePathAndName := fmt.Sprintf("%s/%s", secret.MountPath, secret.Name)

if secret.FilePathAndName != "" {
	filePathAndName = filepath.Join(secret.MountPath, secret.FilePathAndName)
}
```

That should be equivalent to what we have here, and I think with just a single conditional (no `else`) it's easier to reason about: https://play.golang.org/p/-t3Skfz9Yf_k

agent-inject/agent/config_test.go
```diff
@@ -142,6 +142,96 @@ func TestNewConfig(t *testing.T) {
...
+		},
+		{
+			"with volume mount set, relative file path",
+			map[string]string{
+				"vault.hashicorp.com/agent-inject-secret-foo": "db/creds/foo",
+				"vault.hashicorp.com/agent-inject-file-foo":   "nested/foofile",
+				AnnotationVaultSecretVolumePath + "-foo":      "/new/mount/path",
+			},
+			"/new/mount/path/nested/foofile",
+		},
```
catsby, line 201: What is the effective difference between these two scenarios? The `inject-file-foo` value is different by the leading `/` but the both end up in the same file structure:

```
/new/mount/path/special/foofile
```
vs
```
/new/mount/path/nested/foofile
```

agent-inject/agent/annotations.go
```diff
@@ -366,7 +379,14 @@ func (a *Agent) secrets() []*Secret {
...
 
-			secrets = append(secrets, &Secret{Name: name, Path: path, Template: template, Command: command, MountPath: mountPath})
+			var filePathAndName string
+			file := fmt.Sprintf("%s-%s", AnnotationAgentInjectFile, raw)
+
+			if val, ok := a.Annotations[file]; ok {
+				filePathAndName = val
+			}
+
+			secrets = append(secrets, &Secret{Name: name, Path: path, FilePathAndName: filePathAndName, Template: template, Command: command, MountPath: mountPath})
```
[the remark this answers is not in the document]
tvoran: Yep, I agree, refactored a bit in b96d6b4.

agent-inject/agent/config.go
tvoran: Thanks! This is much cleaner. Added in b96d6b4.

agent-inject/agent/config_test.go
tvoran: The leading `/` is the main difference, just making sure both path styles work when a custom mount path is specified. Tried to make the test descriptions a little more clearer in b96d6b4 if that helps.

tvoran:
> I don't quite get the significance of the mentioned absolute pathing (e.g. /some/file vs some/file), they appear to me to produce the same results. 

Yeah, going through this again I think calling out absolute pathing is unnecessary, since both styles just end up inside the secrets volume, i.e. you can't specify an absolute path to write something outside in an arbitrary location within the container. Tried to simplify some of that language in the PR description and code comments in b96d6b4.

PR #164, opened 2020-08-06
feature: Add NAMESPACE (from downward API) to list of environment vars

cyrus-mc:
For the init and sidecar, add NAMESPACE as an environment variable (obtained from downward API).

### DIFF
--- a/agent-inject/agent/container_env.go
+++ b/agent-inject/agent/container_env.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"encoding/base64"
+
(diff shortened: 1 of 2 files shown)


cyrus-mc:
In my specific case the namespace is used to denote the environment and this is useful information to have when writing templates. 

Outside of my specific use case I think it is just good practice to include this information in all PODs.

jasonodonnell:
Bump @cyrus-mc

tomhjp:
Thanks for the PR! Looks like #486 got merged and has this one covered now.

PR #168, opened 2020-08-12
remove beta from v1 resource definition

Kampe:
Should accompany https://github.com/hashicorp/vault-k8s/pull/112 

Most every platform supports > 1.16 now


jasonodonnell:
Hi @Kampe, I'm unsure if we can merge those.  You're right that 1.16 is supported by every platform, but we have many users who aren't using those versions of Kubernetes.  It's my understanding the beta tag is deprecated, but still supported > 1.16.

Thoughts?

elsesiy:
https://github.com/hashicorp/vault-k8s/pull/112#issuecomment-713882263

tvoran:
Thanks for the patch here, though it looks like this was done in #273

Issue #86, opened 2020-02-26
[EKS-IAM Roles for Service Accounts] Possible to Assume Role?


GuillaumeKG:
Hello,

   Some context information:
* I've deployed the Agent Sidecar Injector inside an EKS cluster targeting an external VAULT server. This part works. 
* My pod has a service account which is mapped to an IAM role (https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
* This IAM role is binded to a VAULT role to have access to secret in the Vault  
   Extract of the script:
   ```
   vault write auth/aws/role/$ROLE auth_type=iam bound_iam_principal_arn="$IAM_ROLE_ARN"
   ...
   vault write identity/entity-alias name="$BOUND_IAM_PRINCIPAL_ID" canonical_id=$ENTITY_ID mount_accessor=$ACCESSOR
   ```
Then:
* When I assume this role and try to access to secret via vault kv cli it works.
* When the agent try to connect to the Vault it receives the following error message:
  ```
  [ERROR] auth.handler: error authenticating: error="Error making API request.
  URL: PUT https://vault.hml.services.aws/v1/auth/kubernetes/login
  Code: 400. Errors:
  * missing client token" backoff=1.971367171
  ```

When I use the CLI, I retrieve manually the token with:
````
export VAULT_TOKEN=`vault login -method=aws -token-only role=sdo_spy-test_rw_iam`
```` 
So I suppose, the issue comes from the method `aws` which is not used by the injector by default I guess. 
Is there a setting to change the way the token is retrieved? Or do I need to provide the token manually?

Thanks for your help

Guillaume

**Comment by GuillaumeKG** [comment]
Hi, 
    after re-checking the docs, can I achieve my goal using configuration via configmap? As below:

```
  config.hcl: |
    "auto_auth" = {
      "method" = {
        "config" = {
          "role" = "db-app"
        }
        "type" = "kubernetes"
      }
```
And replace `kubernetes` by `aws`?

I didn't find the doc with all possible fields and values. But I'll try this solution waiting for a feedback.

**Comment by GuillaumeKG** [comment]
Quick update, using the configmap way of configuration works with:
```
    "auto_auth" = {
      "method" = {
        "config" = {
          "type" = "iam"
          "role" = "my-vault-role"
        }
        "type" = "aws"
      }
```
So good, but it uses the ec2 instance role for authentication.  

[...]

PR #169, opened 2020-08-13
Added support for IRSA, fixes - #86

infa-mhadiman:
Adding support for IAM roles for service accounts (IRSA) on EKS. This PR takes care of injecting the volume mounts and env variables when a deployment & SA is configured to use pod identity (IRSA). 

Looks up an app container for aws-iam-token volume mount and if found, uses the same and mounts it to vault init & sidecar containers.
 
**Fixes - #86 **


infa-mhadiman:
Hi @tvoran  ,
I've added changes to resolve the conflicts. Can you please review ? and let me know if there are any further changes.

Thanks

tvoran:
Hi @infa-mhadiman, apologies for the delay, I should be able to try this out this week.

agent-inject/agent/container_env.go
```diff
@@ -42,6 +44,21 @@ func (a *Agent) ContainerEnvVars(init bool) ([]corev1.EnvVar, error) {
...
+			if len(envMap) == 0 || len(envMap) >= 2 {
+				for k, v := range envMap {
+					envs = append(envs, corev1.EnvVar{
+						Name:  k,
+						Value: v,
+					})
+				}
+			} else {
+				log.Println("WARN: Could not find 'AWS ROLE/ AWS_WEB_IDENTITY_TOKEN_FILE' env variables")
+			}
```
tvoran, line 60: I think something like this would be simpler, and we probably don't want to log every time the AWS env vars aren't found:

```suggestion
			for k, v := range envMap {
				envs = append(envs, corev1.EnvVar{
					Name:  k,
					Value: v,
				})
			}
```

If we want to keep the log, I'd suggest changing it to trace or debug level, and it'll need to be a logger based off the hclog in handler.

agent-inject/agent/agent.go
```diff
@@ -521,6 +533,34 @@ func serviceaccount(pod *corev1.Pod) (string, string) {
 	return serviceAccountName, serviceAccountPath
 }
 
+// IRSA support - get aws_iam_token volume mount details to inject to vault containers
+func getAwsIamTokenAccount(pod *corev1.Pod) (string, string) {
```
tvoran, line 537: How about renaming this to getAwsIamTokenVolume?

```suggestion
func getAwsIamTokenVolume(pod *corev1.Pod) (string, string) {
```

agent-inject/agent/agent.go
```diff
@@ -217,22 +226,25 @@ type VaultAgentCache struct {
 // New creates a new instance of Agent by parsing all the Kubernetes annotations.
 func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, error) {
 	saName, saPath := serviceaccount(pod)
+	iamName, iamPath := getAwsIamTokenAccount(pod)
```
tvoran, line 229: Can we call this only if AuthType == "aws"?

tvoran (commented):
Good stuff here! I think it could be really useful, especially now that #213 has been merged.

I left some suggestions, mostly around style. And in general I think the AWS-specific items like this should only be triggered if AuthType=="aws" if possible. Let me know what you think!

infa-mhadiman:
Agreed, I'll make the suggested changes. thanks for review.

infa-mhadiman:
Hey @tvoran,
I've updated the PR with the suggested changes. Can you please review and let me know if there is any change req ?

Thanks

agent-inject/agent/container_env.go
```diff
@@ -59,5 +59,18 @@ func (a *Agent) ContainerEnvVars(init bool) ([]corev1.EnvVar, error) {
...
+	if a.Vault.AuthType == "aws" {
+		envMap := a.getAwsEnvsFromContainer(a.Pod)
+		if len(envMap) == 0 || len(envMap) >= 2 {
+			for k, v := range envMap {
+				envs = append(envs, corev1.EnvVar{
+					Name:  k,
+					Value: v,
+				})
+			}
+		}
```
tvoran, line 72: I don't think we need to check the length of the envMap before iterating here?

```suggestion
		for k, v := range envMap {
			envs = append(envs, corev1.EnvVar{
				Name:  k,
				Value: v,
			})
		}
```

agent-inject/agent/container_env.go
```diff
@@ -2,7 +2,7 @@ package agent
 
 import (
 	"encoding/base64"
-
+
```
tvoran, line 5:
```suggestion

```

agent-inject/agent/container_env.go
infa-mhadiman: Yes, it's not needed thr. I'll remove this.

infa-mhadiman:
This should be good now, can you please review again ? @tvoran

tvoran:
Hi @infa-mhadiman, I was experimenting with IRSA on EKS, and I don't know if something changed recently there, but it looks like the IRSA env vars and token are being added to the vault-agent init and sidecar containers without any special handling required by vault-k8s. I'm wondering if you're still seeing the need for this PR? If so, can you detail your setup/use case?

infa-mhadiman:
hmm, interesting. @tvoran, what is the EKS version ur testing on ?

The pod identity webhook will inject the IRSA env vars and token for all the available containers in a POD. so, wondering how this is possible if the agent injector webhook hasn't mutated the object yet.

During my testing, I observed the mutating webooks invocation to be in the order they existed, the pod identity webhook is being invoked even before, vault injector webhook. So, do you've a vault injector webhook configuration with a different name ? such that it's being invoked before pod identity webhook ?

infa-mhadiman:
Coming to our setup.
we use a old vault 1.3.x on EKS 1.15, 1.16, which does not have OIDC support for vault, the vault sidecar being injected invokes a small program which performs an assume webidentity call, gets the creds and makes it available for vault agent to perform an auth.

infa-mhadiman:
Coming to our setup.
we use a old vault 1.3.x on EKS 1.15, 1.16, which does not have OIDC support for vault, the vault sidecar being injected invokes a small program which performs an assume webidentity call, gets the creds and makes it available for vault agent to perform an auth.

marknet15:
If it helps at my company, we were recently testing with EKS 1.18 & 1.19 and Vault v1.6.1 with the latest version of the Vault sidecar agent to get it up and running with the the web identity IAM service account integration, and we were only seeing the EKS node IAM being used by the Vault sidecar instead of the web identity token. So I think this is definitely still needed.

Our setup makes use of the OIDC IAM provider that maps an IAM role through to the pods service account, which is able to be used fine on the actual pod app container to access AWS resources etc. However we weren't able to authenticate on the vault agent with this same auth as it fell back to the node IAM role.

tvoran:
@infa-mhadiman Yeah, I'm just testing with new EKS clusters, so they're on 1.18, and it certainly wouldn't surprise me if this behavior changed between EKS versions. Nothing special with naming the injector webhook, just the defaults for the chart. 

I do see that the `pod-identity-webhook` in my EKS cluster has `reinvocationPolicy: IfNeeded`, so it'll process a pod again if another webhook modified the pod after the initial mutation. And if I set that as `reinvocationPolicy: Never` I get the behavior you've described. Though with this PR the env variables are set on the vault-agent init container, but the volume mount isn't added. I should be able to do more testing tomorrow.

agent-inject/agent/container_sidecar_test.go
```diff
@@ -99,6 +99,80 @@ func TestContainerSidecarVolume(t *testing.T) {
...
+
+	if err := agent.Validate(); err != nil {
+		t.Errorf("agent validation failed, it shouldn't have: %s", err)
+	}
+
+	container, err := agent.ContainerSidecar()
+
+	// One token volume mount, one config volume mount and two secrets volume mounts
+	require.Equal(t, 5, len(container.VolumeMounts))
+
```
tvoran, line 142: The require.Equal on the next line will also check the length, so I think this check is not needed.

```suggestion
```

tvoran (commented):
The trouble I was having before was because of the check for AnnotationVaultAuthPath instead of AnnotationVaultAuthType when looking for the IAM token volume. With that change it started working for me. Also a few other test comments.

agent-inject/agent/agent.go
```diff
@@ -243,23 +251,29 @@ type VaultAgentCache struct {
 // New creates a new instance of Agent by parsing all the Kubernetes annotations.
 func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, error) {
 	saName, saPath := serviceaccount(pod)
-
+	var iamName, iamPath string
+	if pod.Annotations[AnnotationVaultAuthPath] == "aws" {
```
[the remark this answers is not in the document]
infa-mhadiman: this is a mistake, I overlooked this one. Fixed this. thanks for pointing out.

infa-mhadiman:
@tvoran 
Thanks for the feedback. I've updated with review comments, please review

agent-inject/agent/agent.go
```diff
@@ -588,6 +602,34 @@ func serviceaccount(pod *corev1.Pod) (string, string) {
...
+	}
+	return awsIamTokenAccountName, awsIamTokenAccountPath
+}
+
+// IRSA support - get aws envs to inject to vault containers
+func (a *Agent) getAwsEnvsFromContainer(pod *corev1.Pod) map[string]string {
+	envMap := make(map[string]string)
+	for _, container := range pod.Spec.Containers {
+		for _, env := range container.Env {
+			if strings.Contains(env.Name, "AWS_ROLE_ARN") || strings.Contains(env.Name, "AWS_WEB_IDENTITY_TOKEN_FILE") {
```
tvoran, line 623: I'm curious if this needs to use `.Contains()`? If not, it seems like `==` would be a little tighter:

```suggestion
			if env.Name == "AWS_ROLE_ARN" || env.Name == "AWS_WEB_IDENTITY_TOKEN_FILE" {
```

infa-mhadiman: Okay, Updated. We already know the names. So, we can do a strict comparison.

tvoran (approved):

tvoran:
Thanks for your patience with this one. We'll get it in to the next release!

PR #170, opened 2020-08-18
Add unique home volumes for init/sidecar

jasonodonnell:
This fixes a small bug when a Vault Agent Proxy is used to cache requests from injected agent containers.  The bug is the init container puts a token into $HOME, which is a shared memory volume.  The sidecar doesn't use this token, however, it's sent along with the request so the cache hash is calculated wrong, causing a cache miss.

Each container will now have its own memory volume for $HOME so there is nothing shared between them except the secret volumes.

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -357,7 +357,7 @@ func (a *Agent) Patch() ([]byte, error) {
 	// Add a volume for the token sink
 	a.Patches = append(a.Patches, addVolumes(
 		a.Pod.Spec.Volumes,
-		[]corev1.Volume{a.ContainerTokenVolume()},
+		a.ContainerTokenVolume(),
 		"/spec/volumes")...)
 
 	// Add our volume that will be shared by the containers
--- a/agent-inject/agent/container_volume.go
+++ b/agent-inject/agent/container_volume.go
@@ -60,12 +60,22 @@ func (a *Agent) ContainerVolumes() []corev1.Volume {
 
 // ContainerTokenVolume returns a volume to mount the
 // home directory where the token sink will write to.
-func (a *Agent) ContainerTokenVolume() corev1.Volume {
-	return corev1.Volume{
-		Name: tokenVolumeName,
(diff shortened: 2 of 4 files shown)


agent-inject/agent/container_volume.go
```diff
@@ -60,12 +60,22 @@ func (a *Agent) ContainerVolumes() []corev1.Volume {
...
-func (a *Agent) ContainerTokenVolume() corev1.Volume {
-	return corev1.Volume{
-		Name: tokenVolumeName,
-		VolumeSource: corev1.VolumeSource{
-			EmptyDir: &corev1.EmptyDirVolumeSource{
-				Medium: "Memory",
+func (a *Agent) ContainerTokenVolume() []corev1.Volume {
+	return []corev1.Volume{
+		{
+			Name: tokenVolumeName,
```
tvoran, line 66: I wonder if this should also be named something more distinctive, like `tokenVolumeName + "-sidecar"`?

agent-inject/agent/agent.go
```diff
@@ -357,7 +357,7 @@ func (a *Agent) Patch() ([]byte, error) {
 	// Add a volume for the token sink
 	a.Patches = append(a.Patches, addVolumes(
 		a.Pod.Spec.Volumes,
-		[]corev1.Volume{a.ContainerTokenVolume()},
+		a.ContainerTokenVolume(),
```
tvoran, line 360: I'm not sure if it matters, but I think this will add both the `home` and `home-init` volumes to a pod, even if either the init or sidecar is disabled.

jasonodonnell: It will be available in the `volumes`, but not mounted.  We could make this smarter in the future, but for now it's really not a big deal AFAIK.  Thoughts?

jasonodonnell: Added, was easier than I thought.

agent-inject/agent/container_volume.go
jasonodonnell: I just added two new consts since they're used in the volume structs and mount names.

PR #183, opened 2020-09-28
Add UBI container image

jasonodonnell:
This PR adds UBI flavored container image for Vault K8s.  I notice the Makefile wasn't using a variable created specifically for the image tag, so I swapped it.

The image tag for UBI has `_ubi` appended to it, similar to how we do `ent` images in the Vault project.


tomhjp (approved):
LGTM

sarahethompson (approved):

PR #193, opened 2020-11-14
Modified the DefaultTemplate to only showcase key:value from the prop…

Bioblaze:
…er location.

Currently
data: map[]
metadata: map[created_time:2020-11-14T18:58:09.86960569Z deletion_time: destroyed:false version:1]

is being passed, which makes it incredibly difficult to parse the actual file properly, due to the fact some values could have `Spaces` within them, or even : which would make it alittle harder to properly parse.

Expected Output:

Key:Value
Key:Value
Key:Value

Current Output:
data: map[key:value key:value key:value]
metadata: map[created_time:2020-11-14T18:58:09.86960569Z deletion_time: destroyed:false version:1]

Reason for Edit:

This causes parsing this file to be extremely difficult. Currently I have made a single change adding only a `.data` too the DefaultTemplate. This will allow for a simple Regex Statement, and really easy parsing. 

I Hope this is accepted, <3 Thank you for your time in reading this.

### DIFF
--- a/agent-inject/agent/config.go
+++ b/agent-inject/agent/config.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	DefaultTemplate = "{{ with secret \"%s\" }}{{ range $k, $v := .Data }}{{ $k }}: {{ $v }}\n{{ end }}{{ end }}"
+	DefaultTemplate = "{{ with secret \"%s\" }}{{ range $k, $v := .Data.data }}{{ $k }}: {{ $v }}\n{{ end }}{{ end }}"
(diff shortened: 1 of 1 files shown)


jasonodonnell:
Hi @Bioblaze, thanks for the contribution.

This change will break the default template for all other secret engines because KV2 is the only engine with a nested `.Data.data` structure. Unfortunately this cannot be merged due to this. I would suggest using a custom template to make the data more parseable. You could use the following template in your annotations:

```yaml
vault.hashicorp.com/agent-inject-template-kvv2-secret: |
    {{ with secret "secret/hashicorp" }}{{ range $k, $v := .Data.data }}{{ $k }}: {{ $v }}\n{{ end }}{{ end }}
```

PR #198, opened 2020-12-15
Support for multiple replicas with auto-tls

tvoran:
Adds a `use-leader-elector` option to vault-k8s to enable use of a leader-elector sidecar, which lets multiple replicas of the injector determine a "leader" which is then in charge of generating the CA, Cert, and Key, distributing them to the followers, and patching the webhook caBundle. Defaults to `false`.

The leader-elector sidecar in use is described in detail here: https://kubernetes.io/blog/2016/01/simple-leader-election-with-kubernetes/
The leader-elector containers determine a leader by modifying an Endpoint object, and they expose this information to the injector pod with a local http service that returns the current leader's hostname.

The followers read the cert and key needed for the webhook listener from a k8s Secret, which is updated by the leader when a certificate is near expiration.

### DIFF
--- a/deploy/injector-rbac.yaml
+++ b/deploy/injector-rbac.yaml
@@ -38,3 +38,36 @@ subjects:
 - kind: ServiceAccount
   name: vault-injector
   namespace: vault
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: vault-injector-role
+  labels:
+    app.kubernetes.io/name: vault-injector
+    app.kubernetes.io/instance: vault
+rules:
+- apiGroups: [""]
+  resources: ["endpoints", "secrets"]
+  verbs: 
+    - "create"
+    - "get"
+    - "watch"
+    - "list"
+    - "update"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: vault-injector-rolebinding
+  labels:
+    app.kubernetes.io/name: vault-injector
+    app.kubernetes.io/instance: vault
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: vault-injector-role
+subjects:
+- kind: ServiceAccount
+  name: vault-injector
+  namespace: vault
--- a/helper/cert/source_gen.go
+++ b/helper/cert/source_gen.go
@@ -28,7 +28,7 @@ import (
 
 // Name of the k8s Secret used to share the caBundle between leader and
 // followers
-const certSecretName = "certs"
+const certSecretName = "vault-injector-certs"
 
 // GenSource generates a self-signed CA and certificate pair.
 //
@@ -55,7 +55,7 @@ type GenSource struct {
 	caCertTemplate *x509.Certificate
 	caSigner       crypto.Signer
 
-	K8sClient            *kubernetes.Clientset
+	K8sClient            kubernetes.Interface
 	Namespace            string
 	SecretsCache         informerv1.SecretInformer
 	LeaderElectorEnabled bool
--- a/helper/cert/source_gen.go
+++ b/helper/cert/source_gen.go
@@ -20,9 +20,15 @@ import (
 
 	"github.com/hashicorp/vault-k8s/leader"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	informerv1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
 )
+
+// Name of the k8s Secret used to share the caBundle between leader and
+// followers
+const certSecretName = "certs"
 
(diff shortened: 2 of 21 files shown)


deploy/injector-rbac.yaml
```diff
@@ -38,3 +38,36 @@ subjects:
...
+metadata:
+  name: vault-injector-role
+  labels:
+    app.kubernetes.io/name: vault-injector
+    app.kubernetes.io/instance: vault
+rules:
+- apiGroups: [""]
+  resources: ["endpoints", "secrets"]
+  verbs: 
+    - "create"
```
tomhjp, line 53: Is `create` needed if `deploy/injector-leader-extras.yaml` is deployed? Is this to prevent a race or to allow it to still work if users don't deploy that file?

subcommand/injector/flags.go
```diff
@@ -68,6 +68,9 @@ type Specification struct {
 
 	// TelemetryPath is the AGENT_INJECT_TELEMETRY_PATH environment variable.
 	TelemetryPath string `split_words:"true"`
+
+	// UseLeaderElector is the AGENT_INJECT_USE_LEADER_ELECTOR
```
tomhjp, line 72: nit:

```suggestion
	// UseLeaderElector is the AGENT_INJECT_USE_LEADER_ELECTOR environment variable.
```

helper/cert/source_gen.go
```diff
@@ -17,8 +17,19 @@ import (
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	informerv1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
+// Name of the k8s Secret used to share the caBundle between leader and
+// followers
+const certSecretName = "vault-injector-certs"
```
tomhjp, line 31: This being fixed assumes a maximum of 1 injector deployment per namespace right? Seems reasonable to me

subcommand/injector/command.go
```diff
@@ -175,6 +199,15 @@ func (c *Command) Run(args []string) int {
 	return 0
 }
 
+func getNamespace() string {
+	namespace := os.Getenv("NAMESPACE")
+	if len(namespace) > 0 {
```
tomhjp, line 204: nit: `namespace != ""` seems more canonical?

tomhjp (approved):
Nice! LGTM

helper/cert/source_gen.go
jasonodonnell: There can only be 1 injector deployment because the mutating webhook annotation prefix `vault.hashicorp.com` is not configurable.

deploy/injector-rbac.yaml
tvoran: I think it is, since the leader-elector needs to be able to recreate the Endpoint if it is deleted, and similarly the vault-k8s leader will create the Secret if it doesn't exist when it generates a certificate.

subcommand/injector/flags.go
tvoran: Oops, thanks!

deploy/injector-rbac.yaml
jasonodonnell: @tvoran is correct, this is needed.

subcommand/injector/command.go
tvoran: I'm good with that, updated in 501069e

helper/cert/source_gen.go
```diff
@@ -51,11 +67,26 @@ func (s *GenSource) Certificate(ctx context.Context, last *Bundle) (Bundle, erro
...
 
+	if s.LeaderElector != nil {
+		leaderCheck, err := s.LeaderElector.IsLeader()
+		if err != nil {
+			return result, err
+		}
+		// For followers, run different function here that reads bundle from Secret,
+		// and returns that in the result. That will flow through the existing
+		// notify channel structure, testing if it's the same cert as last, etc.
+		if !leaderCheck {
```
jasonodonnell, line 78: Can we get a log saying whether this pod is the leader or follower? Also wondering if we should apply a label so we can query for the leader (not blocking, future enhancement probably).

Currently I can't tell which is which:

```
[~/Git/vault-agent-demo] kubectl logs vault-injector-57485f744-24jwd -c sidecar-injector
Using leader elector logic
2020-12-16T15:22:43.623Z [INFO]  handler: Starting handler..
Listening on ":8080"...
Updated certificate bundle received. Updating certs...
[~/Git/vault-agent-demo] kubectl logs vault-injector-57485f744-lts8w -c sidecar-injector
Using leader elector logic
2020-12-16T15:22:44.109Z [INFO]  handler: Starting handler..
Listening on ":8080"...
Updated certificate bundle received. Updating certs...
```

jasonodonnell (approved):
Minor feedback but LGTM and tested fine!

helper/cert/source_gen.go
tvoran: Yep, good point, added more logging in ac4b4f4. I set the leader and follower log lines to debug since they are logged pretty frequently.

PR #209, opened 2021-01-13
Store Vault Agent config in a ConfigMap

### LINKED ISSUE #197: VAULT_CONFIG passed as env variable exceeds MAX_ARG_STRLEN
We encountered the following error with the generated sidecar from vault-agent annotations.

`standard_init_linux.go:211: exec user process caused "argument list too long"`

We found two limits: ARG_MAX and MAX_ARG_STRLEN. The latter which defines the max length of an env variable (131072) hurts us. The last successful deployment had a VAULT_CONFIG length of: 122725. After adding more annotations, the new failing sidecar has a VAULT_CONFIG length of 131533. This results in the error above.

We would expect that the injector creates a configmap from the annotations and mounts the configmap into the sidecar.

We know we can use the `vault.hashicorp.com/agent-configmap` annotation and provide the full vault config ourselves, but that requires extra technical  know-how for our developers.

jasonodonnell:
Generated Vault Agent configuration files are stored in a base64 encoded `VAULT_CONFIG` environment variable to avoid managing ConfigMaps in all namespaces where the Injector operates. While this has worked fine, there are limitations on the argument length if the generated configuration file is large enough.

This PR alters the injector to now create, update and delete configmaps with the generated config file and mount it to the pod. The escape hatch annotation `vault.hashicorp.com/agent-configmap` can still be used if users want to make their own configs.

Fixes #197.


agent-inject/agent/annotations.go
```diff
@@ -81,6 +81,11 @@ const (
 	// configuration file and templates can be found.
 	AnnotationAgentConfigMap = "vault.hashicorp.com/agent-configmap"
 
+	// AnnotationAgentGeneratedConfigMapName is the name of the configuration map generated by the
+	// Vault Agent Injector. This is only for internal uses and should not be modified by
+	// users.
+	AnnotationAgentGeneratedConfigMapName = "vault.hashicorp.com/agent-configmap-name"
```
calvn, line 87: If this shouldn't be modified externally, is it possible to not expose this?

jasonodonnell: We add this annotation to the injected pod so we know the name of the configmap generated when we clean it up. We have other annotations (`inject-status`) that are also just used by us. I think this is good to have since a user can check the name of the configmap we've created for them by looking at the annotations. If they change this annotation, then the delete will fail and they will need to clean up the orphaned configmap. I don't see the harm in having this but we could figure out the name of the configmap without it if you think its necessary. This just makes the delete phase much easier.

agent-inject/handler.go
```diff
@@ -124,11 +130,27 @@ func (h *Handler) Mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionRespon
...
+			err := fmt.Errorf("error checking if should delete configmap for agent: %s", err)
+			return admissionError(err)
+		} else if !delete {
+			return resp
+		}
+
+		configMapName := pod.Annotations[agent.AnnotationAgentGeneratedConfigMapName]
+		h.Log.Debug(fmt.Sprintf("deleting configmap %s..", configMapName))
+
+		_ = h.Clientset.CoreV1().ConfigMaps(req.Namespace).Delete(configMapName, &metav1.DeleteOptions{})
```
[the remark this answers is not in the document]
jasonodonnell: Ah yeah, I meant to clean this up but forgot, thanks!

agent-inject/agent/annotations.go
tvoran: It also occurs to me that setting the configmap name in an annotation on the pod will also make things easier for backwards compatibility if we ever decide to change the configmap naming scheme.

calvn: Is it common practice for a particular annotation to be exposed, but not expected to be modified? I am just thinking of possible side-effects like orphaned configmaps like you mentioned.

tomhjp: Unless I'm missing a way in which this example is different, I believe this is fairly common. The k8s API uses annotations to store metadata like previous configuration and modified time etc.

tomhjp: Very much a nit but could a name like `vault.hashicorp.com/agent-configmap-generated-name` make its purpose a bit clearer? It does currently read as though it's an option that should go hand in hand with setting `agent-configmap`.

agent-inject/agent/agent.go
```diff
@@ -355,6 +368,67 @@ func ShouldInject(pod *corev1.Pod) (bool, error) {
...
+	}
+
+	return true, nil
+}
+
+// ConfigMapName determines what the name of the generated configmap should be.
+// When the injector gets a pod creation event, the pod isn't yet created and
+// may not have it's final name (if it's owned by a replicaset) . Here we consider
+// if the pod is owned by a controller or created explicitly by the user.
+func ConfigMapName(pod *corev1.Pod) string {
```
tomhjp, line 421: Does this function need to be exported?

agent-inject/agent/agent.go
```diff
@@ -355,6 +368,67 @@ func ShouldInject(pod *corev1.Pod) (bool, error) {
...
+	}
+
+	// This shouldn't happen so bail.
+	raw, ok = pod.Annotations[AnnotationAgentStatus]
+	if !ok {
+		return false, nil
+	}
+
+	// Previously injected agents
+	if raw != "injected" {
```
tomhjp, line 395: Would it make sense to have a small "enum" type for the possible values of status? IIRC there are only like 2 states, but the definition of those states would make for a nice reference of what the possible states are.

agent-inject/agent/agent.go
```diff
@@ -355,6 +368,67 @@ func ShouldInject(pod *corev1.Pod) (bool, error) {
...
+	if !ok {
+		return false, nil
+	}
+
+	if raw == "" {
+		return false, nil
+	}
+
+	// Only delete config maps if the name of the configmap and the pod are the same.
+	if ConfigMapName(pod) != pod.Annotations[AnnotationAgentGeneratedConfigMapName] {
```
tomhjp, line 410: Under what conditions would be expect to take this path in the logic? If the user has modified the annotation themselves? IIUC it's quite a confusing state to be in, so perhaps it's worth logging something to say "something weird has happened so we'll do the safe thing and leave it alone - note that modifying the agent-configmap-name annotation will cause dangling config maps like this to be left around. The config map with <this> name is _probably_ safe to delete, but check it yourself".

tomhjp (commented):
LGTM modulo Theron's comment on the unhandled error. Just a few additional minor suggestions.

agent-inject/agent/agent.go
jasonodonnell: I think that could be a nice clean up in the future since `shouldInject` is very similar to this.

agent-inject/agent/agent.go
jasonodonnell: Made it private. Thanks!

agent-inject/agent/agent.go
jasonodonnell: @tvoran and I agreed we needed to protect against someone trying to delete a ConfigMap owned by another pod. You could craft a pod name, try a bogus injection and then delete it to trigger a privileged delete on the configmap. This check just ensures this can't be abused.

jasonodonnell:
Some work needed on this so closing and will resubmit when ready.

Issue #205, opened 2020-12-30
Support for injecting HTTPS_PROXY environment variable to the init and sidecar containers


danielfm:
**Is your feature request related to a problem? Please describe.**

It's a best practice to not allow connections to Vault open to all IPs (0.0.0.0/0), and instead allowlist the IP addresses that need access to Vault via ingress rules (firewall in GKE, security group in AWS, etc).

However, in my use case, the injector runs in an old Kubernetes cluster that was created without these security constraints in mind (i.e. Kubernetes nodes with external IPs, clusters with VPC native feature disabled in GKE, etc), so it becomes very difficult to control the egress IPs for injected containers, especially in auto-scaled clusters, which is the case.

**Describe the solution you'd like**

If we were able to configure the injector to set HTTPS_PROXY var to the init and sidecar containers, we would be able to make requests to Vault through a proxy with stable outgoing IP addresses, which makes it much easier to enforce the security rule described above.

This could be configured directly in the injector, so that all injected init/agent sidecars would connect to Vault via the proxy, and/or via annotations, so that we could have more control over which containers communicate to Vault via the proxy.

**Describe alternatives you've considered**

- I'm running in GKE. One alternative would be to recreate clusters with a configuration compatible with [Cloud NAT](https://cloud.google.com/nat/docs/overview) in order to have more control over the outgoing IP address for connections originating from these clusters. This ultimately would allow me to only allowlist traffic from the NAT range. However, in my case, migrating everything away from this cluster would be too much work, and I think I could benefit from Vault's features if I had a way for containers in this cluster to talk to Vault via some sort of proxy

**Additional context**

n/a

jasonodonnell:
@danielfm seems reasonable, we'll take a look!

danielfm:
What do you think would be the best implementation for this? Add the proxy address as an injector command line option?

If you give me some directions regarding the implementation you are more likely to accept, I can send a PR for this.

jasonodonnell:
I think all that needs to be done is adding an annotation `vault.hashicorp.com/https-proxy`, which takes a proxy value. Then set the `HTTPS_PROXY` environment variable on the init and sidecar containers. I _think_ that's all it would take.

jasonodonnell:
Additionally if you want to make this a global value, then yeah a new flag on Vault K8s would also be fine.

danielfm:
Will do. Thanks!

PR #211, opened 2021-01-15
Add support for injecting HTTPS_PROXY env var in agent containers

danielfm:
**Why is this PR necessary, what does it do?**

This PR aims to add to the injector the ability of setting the `HTTPS_PROXY` environment variable in the init/sidecar containers in order to allow requests to Vault to be made via a proxy.

**References:**

Fixes: #205

**Notes:**

If the reviewers feel there are tests missing, please let me know! I may have missed something.

I'll try to run this version of the injector in a test cluster to validate the workflow, but in the mean time, feel free to comment.

### DIFF
--- a/agent-inject/agent/config.go
+++ b/agent-inject/agent/config.go
@@ -30,6 +30,7 @@ type Config struct {
 // Vault contains configuration for connecting to Vault servers
 type VaultConfig struct {
 	Address       string `json:"address"`
+	ProxyAddress  string `json:"proxy_adress"`
 	CACert        string `json:"ca_cert,omitempty"`
 	CAPath        string `json:"ca_path,omitempty"`
 	TLSSkipVerify bool   `json:"tls_skip_verify,omitempty"`
@@ -119,6 +120,7 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		ExitAfterAuth: init,
 		Vault: &VaultConfig{
 			Address:       a.Vault.Address,
+			ProxyAddress:  a.Vault.ProxyAddress,
 			CACert:        a.Vault.CACert,
 			CAPath:        a.Vault.CAKey,
 			ClientCert:    a.Vault.ClientCert,
--- a/agent-inject/agent/config.go
+++ b/agent-inject/agent/config.go
@@ -30,7 +30,6 @@ type Config struct {
 // Vault contains configuration for connecting to Vault servers
 type VaultConfig struct {
 	Address       string `json:"address"`
-	ProxyAddress  string `json:"proxy_adress"`
 	CACert        string `json:"ca_cert,omitempty"`
 	CAPath        string `json:"ca_path,omitempty"`
 	TLSSkipVerify bool   `json:"tls_skip_verify,omitempty"`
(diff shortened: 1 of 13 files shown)


danielfm:
I managed to validate this change in a local Kubernetes cluster.

I validated these flows:

- Without proxy configuration
  - Works as it always did
- With `AGENT_INJECT_PROXY_ADDR` env var set to the injector + no annotations
  - Agent requests are forwarded to Vault via the proxy address specified in the env var
- With `AGENT_INJECT_PROXY_ADDR` env var set to the injector + annotation overwriting the proxy address
  - Agent requests are forwarded to Vault via the proxy address specified in the annotation
- No `AGENT_INJECT_PROXY_ADDR` env var set to the injector + annotation overwriting the proxy address
  - Same as above

Please let me know if you want me to perform any other manual tests.

danielfm:
One question I have is to whether it makes sense to inject the same vale in the `HTTP_PROXY` environment variable, or even allow for different proxies for HTTP and HTTPS.

Since the recommended Vault server setup is to enable TLS, maybe this PR is good enough as is, but it may not work for development clusters without TLS enabled, as requests are sent to Vault in plain HTTP.

Let me know what you think.

agent-inject/agent/config.go
```diff
@@ -119,6 +120,7 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		ExitAfterAuth: init,
 		Vault: &VaultConfig{
 			Address:       a.Vault.Address,
+			ProxyAddress:  a.Vault.ProxyAddress,
```
jasonodonnell, line 123: I was looking in the [Vault documentation but do not see this value](https://www.vaultproject.io/docs/agent). Below the HTTPS_PROXY environment variable is being said, so this seems redundant anyway. Thoughts?

danielfm: I believe this adds support for configuring the proxy address via the configuration file as well. If we choose to include this, we'd have to update that documentation accordingly.

If you don't feel this particular change is necessary, I can drop this and only allow the proxy address configuration to be set via a command line flag and/or the `AGENT_INJECT_PROXY_ADDR` environment variable.

danielfm: We'd also probably tag this field with `omitempty` as well if we choose to maintain it, as this setting is optional.

danielfm: Did I answer your question? Which action do you suggest? 

I can go ahead with your initial suggestion and remove this bit.

jasonodonnell: `AGENT_INJECT_PROXY_ADDR` sets a global default that will be included with every Vault Agent container. If this is not set, the proxy annotation can be used by the user to explicitly set (or override the global).

What I'm trying to clear up is where this setting ultimately lands. Currently its set in two places: the containers environment variable (HTTPS_PROXY) and the Vault Agent configuration file. It should be one or the other, not both. I wasn't able to find references to it in the Vault Agent config settings documentation, so I think the `HTTPS_PROXY` environment variable is the way to go.

danielfm: You are correct. I was confusing the `AgentConfig` struct as being something like an object for carrying the injector configuration, but it's not.

I removed that property in 7cecb13.

jasonodonnell:
All looks good here @danielfm, do you mind resolving the conflicts?

danielfm:
@jasonodonnell done.

jasonodonnell (approved):
LGTM, thanks!

PR #212, opened 2021-01-16
Add annotation to copy volume mounts from existing container

pbar1:
This PR introduces the following annotation: `vault.hashicorp.com/agent-copy-volume-mounts`, whose value is expected to be the name of container (or init container) in the original pod spec whose volume mounts are to be copied into the Vault Agent init/sidecar containers. Volume mounts whose paths include the string `serviceaccount` are ignored so we don't end up copying K8s service account token mounts (ie, `/var/run/secrets/kubernetes.io/serviceaccount`).

The reasoning behind this feature: At Salesforce, we inject all pods with X.509 certs via SPIFFE at runtime that end up in `emptyDir` volumes. We'd like to use the [Vault Agent Auto-Auth Cert Method](https://www.vaultproject.io/docs/agent/autoauth/methods/cert) with these certs; our Vaults already [require mTLS](https://www.vaultproject.io/docs/configuration/listener/tcp#tls_require_and_verify_client_cert) and use the [TLS Certificates Auth Method](https://www.vaultproject.io/docs/auth/cert) against this PKI, so having these volumes mounted to the Vault Agent's init and sidecar containers is required.

Regarding Auto-Auth via the Cert Method - I've got a separate branch that generalizes the injector to support all Auth Methods (#213). We're currently successfully running a build of the injector with these features in our dev environment.

Let me know how I can help to get this merged. Thanks! ☄️

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -128,6 +128,10 @@ type Agent struct {
 	// ExtraSecret is the Kubernetes secret to mount as a volume in the Vault agent container
 	// which can be referenced by the Agent config for secrets. Mounted at /vault/custom/
 	ExtraSecret string
+
+	// CopyVolumeMounts is the name of the container in the Pod whose volume mounts
+	// should be copied into the Vault Agent init and/or sidecar containers.
+	CopyVolumeMounts string
 }
 
 type Secret struct {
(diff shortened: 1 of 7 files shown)


jasonodonnell (commented):
Hi @pbar1, thanks for the contribution, this looks great! I took it for a spin and it works as described and have no comments on the implementation.

One thing I would ask before merging, I think we'll need a handler test case for this. The side effect of this annotation on the agents should result in volume JSON patches, so we'll want to test that is the case. Take a look at the handler test cases and let me know if you have any questions.

pbar1:
Sure, I've just added another test case for the handler, in which the CopyVolumeMounts annotation is specified.

jasonodonnell (approved):
LGTM, thanks!

pbar1:
Thank you!

irsl:
Guys, this is awesome and perfect timing; I just realized I need this for a use case and it was just released yesterday! :)

PR #221, opened 2021-02-09
add xc-prod-image for cross compiling vault-k8s in release pipeline

alvin-huang:
This PR adds a new Make target, `xc-prod-image` which will use `docker buildx` to cross compile the vault-k8s image across 4 different architectures (and optionally publish with `XC_PUBLISH=--push`). This target is used in the release pipeline but could also be called locally if desired and the host machine has all the necessary packages for using `docker buildx` (normal Docker for Mac should _just work_ as long as `DOCKER_CLI_EXPERIMENTAL=enabled`).  The release pipeline changes are found in https://github.com/hashicorp/vault-k8s-release/pull/4. 

This does require us to start cross compiling binaries for the 4 different versions though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 REGISTRY_NAME?=docker.io/hashicorp
 IMAGE_NAME=vault-k8s
 VERSION?=0.8.0
-IMAGE_TAG=$(REGISTRY_NAME)/$(IMAGE_NAME):$(VERSION)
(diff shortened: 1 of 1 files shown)


jasonodonnell (approved):
LGTM but want to get @sarahethompson to sign off as well.

sarahethompson (approved):
LGTM

Issue #84, opened 2020-02-21
Add support for adding templates from a ConfigMap and mounting them as SubPaths


tsurankov:
It would be great if vault-injector could mount rendered templates individually to the particular path inside an application container similar to SubPath for ConfigMap volumes.
For example, if an application requires two different large templates to be rendered and placed at different paths:

- application.properties
- server.xml

Supposing templates could be placed inside a config map along with vault agent config, vault-agent would render and place them on a shared volume like it does now. But then rendered configuration files could be mounted to the particular location inside the container using custom annotation for each of the template files:

```
vault.hashicorp.com/agent-inject-template-properties: application.properties
vault.hashicorp.com/agent-inject-template-path-properties: /apps/conf/application.properties
vault.hashicorp.com/agent-inject-template-server: server.xml
vault.hashicorp.com/agent-inject-template-path-server: /usr/local/tomcat/conf/server.xml
```

Otherwise, additional start up scripts are required to place rendered configurations to their proper locations, what is not really handy

jasonodonnell:
Hi @tsurankov, I may be wrong about what you're requesting, but we already support mounting a ConfigMap with a Vault Agent configuration file and supporting templates.  You could choose where to render the templates this way.

See our documentation for an example: https://www.vaultproject.io/docs/platform/k8s/injector/examples/#configmap-example

dawidmalina:
I agree with @tsurankov that it would be desired to have simple way to define (through annotation) destination of the generated template. Keep in mind @jasonodonnell that not all application can be reconfigured to use path like `/vault/secrets/application.properties` and using configmap as you suggested (even if it possible) is not simple way.

gunnypatel:
@jasonodonnell is there currently a way to support using a template file that isn't embedded into the agent configuration file?  Can we use a template that's already in a configmap?

virtrixke:
Same here. I was also expecting the following example to work (in this case for a sql connection setting in the server.xml file in tomcat). Problem is I don't see an option to redirect the original /vault/secrets/server.xml to /usr/local/tomcat/conf in Tomcat using an environment variable or something so I'm a bit stuck:

```
vault-hashicorp-com/secret-volume-path: "/usr/local/tomcat/conf/"
vault.hashicorp.com/agent-inject-template-server.xml: |
          {{- with secret "secret/<namespace>/<tomcatsecrets>" -}}
          <?xml version='1.0' encoding='utf-8'?>
          .......
[...]

PR #222, opened 2021-02-15
Add annotation to use template path on disk

jfroche:
We currently have two options to define vault agent templates:
either define the template configuration as an inline template in the
annotation or configure the vault agent directly. The former is really
not handy when template is getting complex, the latter forces us to
manage the whole vault agent configuration.

We add a new annotation that enables the vault agent to inject secrets
from a template file on the container disk. Since
https://github.com/hashicorp/vault-k8s/pull/212, this template can be
present in volume defined on the container.

Annotation example:

```yaml
vault.hashicorp.com/agent-inject-secret-foo: 'database/roles/app'
vault.hashicorp.com/agent-inject-template-file-foo: '/etc/my-app/config.toml.tmpl'
vault.hashicorp.com/agent-inject-file-foo: '/etc/my-app/config.toml',
vault.hashicorp.com/agent-copy-volume-mounts: 'MyContainerNameWithVolumes'
```

If a template content is also defined in annotation (using
`vault.hashicorp.com/agent-inject-template`, the template on disk won't be used.

refs #84

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -45,6 +45,15 @@ const (
 	// If not provided, a default generic template is used.
 	AnnotationAgentInjectTemplate = "vault.hashicorp.com/agent-inject-template"
 
+	// AnnotationAgentInjectTemplateFile is the optional key annotation that configures Vault
+	// Agent what template on disk to use for rendering the secrets.  The name
+	// of the template is any unique string after "vault.hashicorp.com/agent-inject-template-file-",
+	// such as "vault.hashicorp.com/agent-inject-template-file-foobar".  This should map
+	// to the same unique value provided in "vault.hashicorp.com/agent-inject-secret-".
+	// The value is the filename and path of the template used by the agent to render the secrets.
+	// If not provided, the template content key annotation is used.
+	AnnotationAgentInjectTemplateFile = "vault.hashicorp.com/agent-inject-template-file"
+
 	// AnnotationAgentInjectToken is the annotation key for injecting the token
 	// from auth/token/lookup-self
 	AnnotationAgentInjectToken = "vault.hashicorp.com/agent-inject-token"
@@ -378,6 +387,12 @@ func (a *Agent) secrets() []*Secret {
 			templateName := fmt.Sprintf("%s-%s", AnnotationAgentInjectTemplate, raw)
 			if val, ok := a.Annotations[templateName]; ok {
 				s.Template = val
+			}
+			if s.Template == "" {
+				templateFileAnnotation := fmt.Sprintf("%s-%s", AnnotationAgentInjectTemplateFile, raw)
+				if val, ok := a.Annotations[templateFileAnnotation]; ok {
+					s.TemplateFile = val
+				}
 			}
 
 			s.MountPath = a.Annotations[AnnotationVaultSecretVolumePath]
--- a/agent-inject/agent/annotations_test.go
+++ b/agent-inject/agent/annotations_test.go
@@ -473,6 +473,137 @@ func TestTemplateShortcuts(t *testing.T) {
 	}
 }
 
+func TestSecretMixedTemplatesAnnotations(t *testing.T) {
+	tests := []struct {
+		annotations     map[string]string
+		expectedSecrets map[string]Secret
+	}{
+		{
+			map[string]string{
+				"vault.hashicorp.com/agent-inject-secret-foobar":        "test1",
+				"vault.hashicorp.com/agent-inject-template-foobar":      "",
+				"vault.hashicorp.com/agent-inject-template-file-foobar": "/etc/config.tmpl",
+				"vault.hashicorp.com/agent-inject-secret-test2":         "test2",
+				"vault.hashicorp.com/agent-inject-template-test2":       "foobarTemplate",
+				"vault.hashicorp.com/agent-inject-template-file-test2":  "",
+			},
+			map[string]Secret{
+				"foobar": Secret{
+					Name:         "foobar",
+					Path:         "test1",
+					Template:     "",
+					TemplateFile: "/etc/config.tmpl",
+					MountPath:    secretVolumePath,
+				},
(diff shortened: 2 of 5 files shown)


jfroche:
Here is a working example in OpenShift with the vault agent template stored in a `ConfigMap`:

```
$ cat database-config.txt.tpl
{{ with secret "kv_cirb_cicd_openshift/data/testapp-dev/external-service" -}}
ext-username={{ .Data.data.username }}
ext-password={{ .Data.data.password }}
{{ end -}}

{{ with secret "kv_cirb_cicd_openshift/data/testapp-dev/dbcreds" -}}
username={{ .Data.data.username }}
password={{ .Data.data.password }}
{{ end -}}

$ oc create configmap database-cfg --from-file=database-config.txt.tpl
```

We create the `Deployment` with annotation referencing the previous `ConfigMap`:
```
$ cat testapp-deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
  labels:
    app: testapp
  name: testapp
  namespace: testapp-dev
spec:
  replicas: 1
  selector:
    matchLabels:
      app: testapp
  template:
    metadata:
      annotations:
        vault.hashicorp.com/agent-copy-volume-mounts: testapp 
        vault.hashicorp.com/agent-inject-secret-database-config.txt: "" 
        vault.hashicorp.com/agent-inject-template-source-database-config.txt: /vault/config/database.txt.tpl 
        vault.hashicorp.com/auth-path: auth/openshift-sbx
        vault.hashicorp.com/ca-cert: /vault/tls/ca.crt
        vault.hashicorp.com/role: ocp-sbx-testapp
        vault.hashicorp.com/tls-secret: vault-sta-ca-crt
      labels:
        app: testapp
    spec:
      containers:
      - image: registry.redhat.io/rhscl/httpd-24-rhel7:latest
        name: testapp
        resources:
          limits:
            memory: 1G
          requests:
            cpu: 100m
        volumeMounts:
        - mountPath: /vault/config
          name: database-cfg
      volumes:
      - configMap:
          defaultMode: 420
          name: database-cfg
        name: database-cfg
$ oc apply -f testapp-deployment.yaml
deployment.apps/testapp created

$ oc get pods
NAME                        READY   STATUS    RESTARTS   AGE
testapp-5bfc4d7954-mwvsl   2/2     Running   0          87s
```

We can see the `ConfigMap` template being interpolated and written in the destination volume:
```
$ oc rsh -c testapp testapp-5bfc4d7954-mwvsl cat /vault/secrets/database-config.txt
ext-username=ext-user
ext-password=pwd-ext-123

username=user123
password=pass123
```

darkebe:
We really need this good feature for big config file.

agent-inject/agent/annotations.go
```diff
@@ -45,6 +45,15 @@ const (
...
 	AnnotationAgentInjectTemplate = "vault.hashicorp.com/agent-inject-template"
 
+	// AnnotationAgentInjectTemplateSource is the optional key annotation that configures Vault
+	// Agent what template on disk to use for rendering the secrets.  The name
+	// of the template is any unique string after "vault.hashicorp.com/agent-inject-template-source-",
+	// such as "vault.hashicorp.com/agent-inject-template-source-foobar".  This should map
+	// to the same unique value provided in "vault.hashicorp.com/agent-inject-secret-".
+	// The value is the filename and path of the template used by the agent to render the secrets.
+	// If not provided, the template content key annotation is used.
+	AnnotationAgentInjectTemplateSource = "vault.hashicorp.com/agent-inject-template-source"
```
jasonodonnell, line 55: Thoughts on making this `vault.hashicorp.com/agent-inject-template-file` instead?

jasonodonnell:
His @jfroche! This looks great and is a good feature. I had one small suggestion on the name of the annotation, but overall it looks good to me. By the way, there's a new annotation that allows you to mount additional secrets `vault.hashicorp.com/agent-extra-secret` which could be used to mount the template sources.

agent-inject/agent/annotations_test.go
```diff
@@ -473,6 +473,66 @@ func TestTemplateShortcuts(t *testing.T) {
...
 
+func TestSecretTemplateSourceAnnotations(t *testing.T) {
+	tests := []struct {
+		annotations            map[string]string
+		expectedKey            string
+		expectedTemplate       string
+		expectedTemplateSource string
+	}{
+		{
+			map[string]string{
```
jasonodonnell, line 484: I think test coverage on a mixed case would be good to have. For example, using a template file for a large template but using a custom template via annotation for another.

agent-inject/agent/annotations.go
jfroche: Indeed this annotation is clearer. I have updated the code.

agent-inject/agent/annotations_test.go
jfroche: I have added an extra test for this mixed case.

jasonodonnell: Thanks!

jasonodonnell (approved):
LGTM, thanks!

norman-zon:
> By the way, there's a new annotation that allows you to mount additional secrets `vault.hashicorp.com/agent-extra-secret` which could be used to mount the template sources.

I don't understand the decision for creating this annotation, but not `vault.hashicorp.com/agent-extra-configmap`.
You say `...extra-secret` could be used to mount the template source. But the template does not contain sensitive data, the sensitive data is stored in the secret I reference with `vault.hashicorp.com/agent-inject-secret`. So for me it seems logical to add an annotation `vault.hashicorp.com/agent-extra-configmap` that can mount a config map and offer the advantage of being able to handle large templates in clear text.
I understand I could achieve the same using `vault.hashicorp.com/agent-copy-volume-mounts`, but exposing all volumes of my container to the agent container just doesn't feel like the right way to do this. The same applies to the method shown on the "ConfigMap Example" in the docs. Having to duplicate the template for `config.hcl` and `config-init.hcl` seems very error prone to me.

Issue #68, opened 2020-02-04
usage of injector in deployments defined in helm chart


dbeal-wiser:
when defining an injection template annotation within a deployment defined within a helm chart, the following syntax is required due to golang templates using the same syntax (using one of the examples from the website):

```
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: app-example
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-example-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: app-example
  template:
    metadata:
      labels:
        app: app-example
      annotations:
        vault.hashicorp.com/agent-inject: "true"
        vault.hashicorp.com/agent-inject-secret-db-creds: "database/creds/db-app"
        vault.hashicorp.com/agent-inject-template-db-creds: |
          # for helm we have to escape golang templates to use golang templates for vault
          {{ printf `{{- with secret \"database/creds/db-app\" -}}
          postgres://{{ .Data.username }}:{{ .Data.password }}@postgres:5432/appdb?sslmode=disable
          {{- end }}` }}
        vault.hashicorp.com/role: "db-app"
        vault.hashicorp.com/ca-cert: "/vault/tls/ca.crt"
        vault.hashicorp.com/client-cert: "/vault/tls/client.crt"
        vault.hashicorp.com/client-key: "/vault/tls/client.key"
        vault.hashicorp.com/tls-secret: "vault-tls-client"
    spec:
      containers:
        - name: app
          image: "app:1.0.0"
      serviceAccountName: app-example
```

this syntax does not seem immediately obvious to me, and I understand it would be difficult to change. Could a documentation example/info be added to cover info about using vault inject templates being used inside a helm chart?

jasonodonnell:
Hi @dbeal-wiser, this should work:

```
   annotations:  
        vault.hashicorp.com/agent-inject-template-account: |
          {{` {{- with secret "path/to/secret" -}}
          {{ .Data.test }}
          {{- end }} `}}
```

hayderimran7:
@jasonodonnell  follow-up question, what if `"path/to/secret"` comes from values.yaml 
i want to use something like:
```
   annotations:  
        vault.hashicorp.com/agent-inject-template-account: |
          {{` {{- with secret .Values.vault.path  -}}
          {{ .Data.test }}
          {{- end }} `}}
```

hayderimran7:
got it working 🎆 
```
[...]

PR #226, opened 2021-03-03
Add annotations for customizing template delimiters

caitlinelfring:
Currently, you can override the default delimiters with the vault injector by defining a config file via a configmap https://www.vaultproject.io/docs/agent/template#left_delimiter. That isn't always an ideal way to handle templating a secret, especially when using Helm

There is some difficulty when using the vault injector template via annotations, especially when using it in conjunction with Helm. See #68

This PR adds functionality to define the left and right delimiters via annotations.


sudomann:
Is this still going to be added?

tvoran:
Added in #664

PR #229, opened 2021-03-10
Enable persistent agent caching

tvoran:
Adds support for persistent agent caching introduced in https://github.com/hashicorp/vault/pull/10938

Controlled by the "vault.hashicorp.com/agent-cache-enable" annotation, with a few caveats:

If "vault.hashicorp.com/agent-cache-enable": "true",
* ...and if init and sidecar enabled, persistent caching enabled
* ...and if sidecar only, just memory caching without persistence enabled
* ...and if init only, no caching at all

When persistent caching is enabled, a `vault-agent-cache` memory volume is mounted at `/vault/agent-cache` on the init and sidecar containers.

Also adds an annotation "vault.hashicorp.com/agent-cache-exit-on-err", which corresponds to the exit_on_err config setting introduced in https://github.com/hashicorp/vault/pull/10938. Controls whether the agent will exit on an error while restoring the persistent cache.

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -27,6 +27,7 @@ const (
 	DefaultAgentCacheEnable              = "false"
 	DefaultAgentCacheUseAutoAuthToken    = "true"
 	DefaultAgentCacheListenerPort        = "8200"
+	DefaultAgentCacheExitOnErr           = false
 	DefaultAgentUseLeaderElector         = false
 )
 
@@ -230,6 +231,10 @@ type VaultAgentCache struct {
 
 	// UseAutoAuthToken configures whether the auto auth token is used in cache requests
 	UseAutoAuthToken string
+
+	// ExitOnErr configures whether the agent will exit on an error while
+	// restoring the persistent cache
+	ExitOnErr bool
 }
 
 // New creates a new instance of Agent by parsing all the Kubernetes annotations.
@@ -335,10 +340,16 @@ func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, erro
 		return agent, err
 	}
 
+	agentCacheExitOnErr, err := agent.agentCacheExitOnErr()
+	if err != nil {
+		return agent, err
+	}
+
 	agent.VaultAgentCache = VaultAgentCache{
 		Enable:           agentCacheEnable,
 		ListenerPort:     pod.Annotations[AnnotationAgentCacheListenerPort],
 		UseAutoAuthToken: pod.Annotations[AnnotationAgentCacheUseAutoAuthToken],
+		ExitOnErr:        agentCacheExitOnErr,
 	}
 
 	return agent, nil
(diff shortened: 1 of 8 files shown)


agent-inject/agent/agent.go
```diff
@@ -335,11 +343,18 @@ func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, erro
...
+		return agent, err
+	}
+
 	agent.VaultAgentCache = VaultAgentCache{
 		Enable:           agentCacheEnable,
 		ListenerPort:     pod.Annotations[AnnotationAgentCacheListenerPort],
 		UseAutoAuthToken: pod.Annotations[AnnotationAgentCacheUseAutoAuthToken],
+		ExitOnErr:        agentCacheExitOnErr,
 	}
+	agent.VaultAgentCache.Persist = agent.agentCachePersist()
```
calvn, line 357:
```suggestion
		ExitOnErr:        agentCacheExitOnErr,
		Persist:          agent.agentCachePersist(),
	}
```

agent-inject/agent/annotations.go
```diff
@@ -550,6 +558,22 @@ func (a *Agent) agentCacheEnable() (bool, error) {
 	return strconv.ParseBool(raw)
 }
 
+func (a *Agent) agentCachePersist() bool {
```
calvn, line 561: Nit: the caller is `*Agent` in here so you could rename this to `cachePersist` (same with the func below).

calvn (approved):
Not super familiar with the codebase yet, but looks good! Left some smaller suggestions.

agent-inject/agent/annotations.go
```diff
@@ -550,6 +558,22 @@ func (a *Agent) agentCacheEnable() (bool, error) {
...
+	if a.VaultAgentCache.Enable && a.PrePopulate && !a.PrePopulateOnly {
```
calvn, line 562: Determining whether we need persist like so may be fine for now, though I do wonder if we'd have to expose this through annotations down the road.

agent-inject/agent/config_test.go
```diff
@@ -306,8 +308,8 @@ func TestConfigVaultAgentCache(t *testing.T) {
 		t.Error("agent Cache should be enabled")
 	}
 
-	if config.Cache.UseAuthAuthToken != "force" {
-		t.Errorf("agent Cache use_auto_auth_token should be 'force', got %s instead", config.Cache.UseAuthAuthToken)
+	if config.Cache.UseAutoAuthToken != "force" {
```
jasonodonnell, line 311: Nice catch.

jasonodonnell (approved):
LGTM!

agent-inject/agent/agent.go
tvoran: Yep, I reworked this a bit in 00f91529016d93f037dde237a0897e98500cab96 to set it all in the struct, since `agentCachePersist()` also checks whether the cache is enabled.

agent-inject/agent/annotations.go
tvoran: Yep, done in 00f91529016d93f037dde237a0897e98500cab96

tvoran:
FWIW, I also tested these changes with the current version of vault-agent (1.6.3), and the extra persist cache config is ignored without errors. 

So for someone using a pre-1.7 vault-agent, the main difference is the init container's vault-agent will also have caching enabled (as opposed to just the sidecar), but pre-1.7 this is just memory cache, so it shouldn't change behavior for the user.

PR #235, opened 2021-03-19
allow cpu/mem requests and limits to be set globally in the injector

### LINKED ISSUE #166: Custom Vault agent container resource defaults
**Is your feature request related to a problem? Please describe.**
Currently there is no way to specify default resource values for Vault Agent containers created by the agent-injector webhook. These defaults have fixed values and can only be overridden via annotations on the resource being mutated.

**Describe the solution you'd like**
It would be useful to be able to provide default resource requests/limits that will apply to Vault Agent init/sidecar containers added to pods via the webhook. These defaults could then be specified in the [agent-injector Helm chart](https://github.com/hashicorp/vault-helm/blob/master/values.yaml#L26).

**Describe alternatives you've considered**
For now, using the `vault.hashicorp.com/agent-requests-`/`vault.hashicorp.com/agent-limits-` annotations per deployment works, but it would be helpful to be able to move this configuration to the agent-injector.

### LINKED ISSUE #216: Change defaults for agent sidecar requests and limits
The agent sidecar and init container are quite lightweight in my experience.  Because of this, the defaults for requests and limits are quite high.  Rather than have to change them on every workload that we annotate. it would be great if there were a way to pass in values to the vault-agent-injector operator so that it set the requests and limits to a configurable default rather than require each annotated workload contain the same annotations.

I believe that the code in question is defined here: https://github.com/hashicorp/vault-k8s/blob/master/agent-inject/agent/container_sidecar.go#L14-L17

As it stands, our metrics show that regardless of the number of secrets a workload is utilizing, nor how often they are refreshing, the CPU of the sidecar never goes above 15m, and mostly hovers between 1m-5m.  This is DRASTICALLY lower than the default of 250/500.  Same with memory, we never pass 15-20Mi, but the defaults are set to 64/128.

Without adding 4 annotations to each workload (sometimes people forget and us operators are left to pick up the pieces), we end up getting a cluster that is likely over provisioned due to autoscaling, or we get CPUOvercommit alerts from prometheus.

When using the helm chart, I would expect to see values like this:

```
injector:
  sidecar:
    resources: # Sets the default resource requests/limits for the sidecar container.  Can still be overridden with annotations
      limits:
        cpu: 50m
        memory: 32Mi
      requests:
        cpu: 25m
        memory: 16Mi
```

joemiller:
Adds new flags and env vars to the injector config that allow changing the cpu/mem
requests and limits of the injected containers:

- `-cpu-request` / `AGENT_INJECT_CPU_REQUEST`
- `-cpu-limit` / `AGENT_INJECT_CPU_LIMIT`
- `-mem-request` / `AGENT_INJECT_MEM_REQUEST`
- `-mem-limit` / `AGENT_INJECT_MEM_LIMIT`

closes #216 
closes #166

### DIFF
--- a/deploy/injector-deployment.yaml
+++ b/deploy/injector-deployment.yaml
@@ -76,6 +76,14 @@ spec:
               value: "vault-agent-injector-svc,vault-agent-injector-svc.$(NAMESPACE),vault-agent-injector-svc.$(NAMESPACE).svc"
             - name: AGENT_INJECT_USE_LEADER_ELECTOR
               value: "true"
+            - name: AGENT_INJECT_CPU_REQUEST
+              value: "250m"
+            - name: AGENT_INJECT_MEM_REQUEST
+              value: "64Mi"
+            - name: AGENT_INJECT_CPU_LIMIT
+              value: "500m"
+            - name: AGENT_INJECT_MEM_LIMIT
+              value: "128Mi"
           args:
             - agent-inject
             - 2>&1
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -227,6 +227,10 @@ type AgentConfig struct {
 	SameID             bool
 	SetSecurityContext bool
 	ProxyAddress       string
+	ResourceRequestCPU string
+	ResourceRequestMem string
+	ResourceLimitCPU   string
+	ResourceLimitMem   string
 }
 
 // Init configures the expected annotations required to create a new instance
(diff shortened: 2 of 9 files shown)


joemiller:
Is there documentation I should update as well?
thanks!

jasonodonnell:
@joemiller I will update the documentation, thank you!

jasonodonnell (commented):
Hi @joemiller, thanks for the contribution! This looks great but can you add these environment variables to the `deploy/deployment.yaml` file?

jasonodonnell:
Once that's added I'll approve this and merge @joemiller. Thanks again!

joemiller:
@jasonodonnell updated the `deploy/injector-deployment.yaml` with the 4 new env vars

deploy/injector-deployment.yaml
```diff
@@ -76,6 +76,14 @@ spec:
               value: "vault-agent-injector-svc,vault-agent-injector-svc.$(NAMESPACE),vault-agent-injector-svc.$(NAMESPACE).svc"
             - name: AGENT_INJECT_USE_LEADER_ELECTOR
               value: "true"
+            - name: AGENT_INJECT_CPU_REQUEST
```
jasonodonnell, line 79: Would you mind setting these values to the backend defaults? These defaults have been here for a while and don't want to change the existing behavior.

```
	DefaultResourceLimitCPU   = "500m"
	DefaultResourceLimitMem   = "128Mi"
	DefaultResourceRequestCPU = "250m"
	DefaultResourceRequestMem = "64Mi"
```

joemiller: oh my mistake, that was my intent actually, but I was looking at the values in one of the unit tests and didn't realize they were the non-defaults. Updating now!

joemiller: @jasonodonnell done, updated

jasonodonnell (approved):
LGTM, thanks!

PR #242, opened 2021-04-02
Add flag and annotation to override default template

### LINKED ISSUE #116: [Feature] Allow altering consul template for default secret output
Hi Vault mavericks!

I like to be able to change the default output format at the cluster level. I did not see such an option, if there is can you please direct me to it?

Thank you.

jasonodonnell:
The current default template (a map) is not easy to consume by applications due to the weird formatting. I've added a global default via flag/env (`AGENT_INJECT_DEFAULT_TEMPLATE`), as well as an annotation (`agent-inject-default-template`) to override this to either be `map` or `json`. I'm keeping the default as `map` so this is backwards compatible in case any application is actually using that style of value.

Once merged we can expose the global in Vault Helm.

Fixes #116.

### DIFF
--- a/agent-inject/agent/container_sidecar_test.go
+++ b/agent-inject/agent/container_sidecar_test.go
@@ -40,8 +40,11 @@ func TestContainerSidecarVolume(t *testing.T) {
 
 	pod := testPod(annotations)
 	var patches []*jsonpatch.JsonPatchOperation
+	agentConfig := AgentConfig{"foobar-image", "http://foobar:1234", DefaultVaultAuthType,
+		"test", "test", true, "1000", "100", DefaultAgentRunAsSameUser,
+		DefaultAgentSetSecurityContext, "", "map"}
 
-	err := Init(pod, AgentConfig{"foobar-image", "http://foobar:1234", DefaultVaultAuthType, "test", "test", true, "1000", "100", DefaultAgentRunAsSameUser, DefaultAgentSetSecurityContext, ""})
+	err := Init(pod, agentConfig)
 	if err != nil {
 		t.Errorf("got error, shouldn't have: %s", err)
 	}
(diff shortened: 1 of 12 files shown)


agent-inject/agent/container_sidecar_test.go
```diff
@@ -109,10 +109,11 @@ func TestContainerSidecar(t *testing.T) {
 
 	agentConfig := AgentConfig{
 		"foobar-image", "http://foobar:1234", DefaultVaultAuthType, "test", "test", false, "1000", "100",
-		DefaultAgentRunAsSameUser, DefaultAgentSetSecurityContext, "https://proxy:3128",
+		DefaultAgentRunAsSameUser, DefaultAgentSetSecurityContext, "https://proxy:3128", "map",
 		DefaultResourceRequestCPU, DefaultResourceRequestMem, DefaultResourceLimitCPU, DefaultResourceLimitMem,
 	}
-	err := Init(pod, agentConfig)
+
+  err := Init(pod, agentConfig)
```
tvoran, line 116: This indentation looks off.

agent-inject/agent/container_sidecar_test.go
```diff
@@ -281,10 +283,11 @@ func TestContainerSidecarConfigMap(t *testing.T) {
 
 	agentConfig := AgentConfig{
 		"foobar-image", "http://foobar:1234", DefaultVaultAuthType, "test", "test", true, "1000", "100",
-		DefaultAgentRunAsSameUser, DefaultAgentSetSecurityContext, "",
+		DefaultAgentRunAsSameUser, DefaultAgentSetSecurityContext, "", "map",
 		DefaultResourceRequestCPU, DefaultResourceRequestMem, DefaultResourceLimitCPU, DefaultResourceLimitMem,
 	}
-	err := Init(pod, agentConfig)
+
+  err := Init(pod, agentConfig)
```
tvoran, line 290: This indentation looks off.

subcommand/injector/flags.go
```diff
@@ -132,6 +135,8 @@ func (c *Command) init() {
 		"Path under which to expose metrics")
 	c.flagSet.BoolVar(&c.flagUseLeaderElector, "use-leader-elector", agent.DefaultAgentUseLeaderElector,
 		fmt.Sprintf("Use leader elector to coordinate multiple replicas when updating CA and Certs with auto-tls"))
+	c.flagSet.StringVar(&c.flagDefaultTemplate, "default-template", "map",
+		"Sets the default template type (map or json). Defaults to map.")
```
tvoran, line 139: Should the default be referencing a variable like the other flags do? i.e.

```suggestion
	c.flagSet.StringVar(&c.flagDefaultTemplate, "default-template", agent.DefaultTemplate,
		fmt.Sprintf("Sets the default template type (map or json). Defaults to %s.", agent.DefaultTemplate)
```

tvoran (approved):
Nice! Worked for me when I tested it. Just a few notes.

agent-inject/agent/container_sidecar_test.go
jasonodonnell: Fixed.

subcommand/injector/flags.go
jasonodonnell: Fixed.

agent-inject/agent/container_sidecar_test.go
jasonodonnell: Fixed.

Issue #207, opened 2021-01-06
Newline in TokenTemplate breaks token usability in some cases


jeremiahsnapp:
**Describe the bug**

The `\n` in [TokenTemplate](https://github.com/hashicorp/vault-k8s/blob/v0.7.0/agent-inject/agent/config.go#L12) breaks token usability in some cases. We have an application which reads the contents of the token file and uses it with the [vault ruby client](https://github.com/hashicorp/vault-ruby) to make API requests. The `\n` in the retrieved token value broke the API request and the vault ruby client returned the message `header field value cannot include CR/LF`.

Also, appending a `\n` to the token file is different behavior when compared to the way vault cli creates a `~/.vault-token` token file and the way vault agent writes the token to a `sink` file.

Can the `\n` be removed from `TokenTemplate`?

tvoran:
Thanks for the report here! The newline in the injected token was removed as part of #250 since it's now injected using the same sink mechanism in vault-agent.

Issue #195, opened 2020-11-17
Error on Token Injection


sidewinder12s:
**Describe the bug**
It appears that when the injector is set to inject the Vault Token into the running pod, it throws a somewhat confusing error. I don't think it actually impacts functionality as I appear to still have the token.

**To Reproduce**

1. Deploy a sample app with the token injection behavior

Application deployment:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sidecar-test
  namespace: default
  labels:
    app: sidecar-test
spec:
  selector:
    matchLabels:
      app: sidecar-test
  replicas: 1
  template:
    metadata:
      annotations:
        vault.hashicorp.com/agent-inject: "true"
        vault.hashicorp.com/role: "default"
        vault.hashicorp.com/agent-inject-token: "true" # Inject the Token at /vault/secrets/token
      labels:
        app: sidecar-test
    spec:
      serviceAccountName: default
      containers:
        - name: sidecar-test
          image: jweissig/app:0.0.1
          env:
            - name: "VAULT_TOKEN" # Expose the Vault Token in the container on the common environment variable
              value: "$(cat /vault/secrets/token)"
```
```log
==> Vault agent started! Log data will stream in below:

==> Vault agent configuration:

                     Cgo: disabled
               Log Level: info
                 Version: Vault v1.5.4
             Version Sha: 1a730771ec70149293efe91e1d283b10d255c6d1

2020-11-17T00:47:46.752Z [INFO]  sink.file: creating file sink
2020-11-17T00:47:46.752Z [INFO]  sink.file: file sink configured: path=/home/vault/.vault-token mode=-rw-r-----
2020-11-17T00:47:46.753Z [INFO]  template.server: starting template server
2020/11/17 00:47:46.753295 [INFO] (runner) creating new runner (dry: false, once: false)
2020/11/17 00:47:46.753820 [INFO] (runner) creating watcher
2020-11-17T00:47:46.814Z [INFO]  sink.server: starting sink server
2020-11-17T00:47:46.814Z [INFO]  auth.handler: starting auth handler
2020-11-17T00:47:46.814Z [INFO]  auth.handler: authenticating
2020-11-17T00:47:51.855Z [INFO]  auth.handler: authentication successful, sending token to sinks
2020-11-17T00:47:51.855Z [INFO]  auth.handler: starting renewal process
2020-11-17T00:47:51.855Z [INFO]  sink.file: token written: path=/home/vault/.vault-token
[...]

PR #250, opened 2021-04-21
Use a file sink for agent-inject-token

### LINKED ISSUE #237: Token inject annotation does not work when agent cache is enabled
* Vault server: Vault 1.6.2, installed with hashicorp/vault 0.9.1 helm chart
* The k8s injector is configured to use hashicorp/vault-k8s:0.9.0 (`injector.image.tag=0.9.0`)

Pod annotations used:

    vault.hashicorp.com/agent-image: vault:1.7.0-rc1
    vault.hashicorp.com/agent-cache-enable: "true"
    vault.hashicorp.com/role: "bugreport"
    vault.hashicorp.com/agent-inject-token: "true"
    vault.hashicorp.com/agent-pre-populate: "true"
    vault.hashicorp.com/agent-pre-populate-only: "false"

After injection, the `/vault/secrets/token` file has contents `<no value>`, which is the output generated by [Consul Template](https://github.com/hashicorp/consul-template/blob/master/docs/templating-language.md#secret) when a map key does not exist. The [agent cache listener removed the `id` token from the  `auth/token/lookup-self` response](https://github.com/hashicorp/vault/blob/master/command/agent/cache/handler.go#L116).

I found this behavior of the agent cache listener to be a little unexpected (and undocumented). If it's intended behavior, then the k8s annotations should probably throw an error or warn when `agent-cache` is enabled with `inject-token`.

---

## The fix / work-around I'm using

I'd like to use caching, so as a work-around I've removed the `vault.hashicorp.com/agent-inject-token` annotation, and instead I use the following:

      {{- with secret "auth/token/lookup-self" -}}
        {{- file "/home/vault/.vault-token" -}}
      {{- end -}}

tvoran:
Since injecting the token from `auth/token/lookup-self` as a secret doesn't work when persistent caching is enabled, this PR uses a file sink instead to make the auto-auth token available to the app container at `/vault/secrets/token`. This also means ACLs are no longer required for `auth/token/lookup-self`. It also gets around the warning noted in #195 and the newline issue in #207.

Fixes #237, #195, #207.


calvn (approved):

tomhjp (approved):
LGTM

jasonodonnell (approved):
LGTM!

PR #251, opened 2021-04-22
Add annotation support for multiple ConfigMap templates

### LINKED ISSUE #102: Support for multiple Vault agent configmaps in injector annotations
Feature req: Vault agent injector "vault.hashicorp.com/agent-configmap" annotation supports multiple configmap names (comma separated) 

Right now, the configmaps referenced by annotation must provide configs called config.hcl and config-init.hcl. This means that even if the annotation string included multiple configmaps, the files would step on each-other as they go directly to /vault/configs/config.hcl and /vault/configs/config-init.hcl.

Possible solution: 
- "vault.hashicorp.com/agent-configmap" split into "vault.hashicorp.com/init-agent-configmap" and "vault.hashicorp.com/agent-configmap" to remove dependency on naming for determining which is used for init-containers vs sidecars. 
- config hcl files go into subdirectories of /vault/configs/init and /vault/configs/sidecar to avoid filename collision
- `vault agent` command updated to use all .hcl configs under /vault/configs/init or /vault/configs/sidecar as appropriate

jrcasso:
If applied, this pull request will create a new annotation, `"vault.hashicorp.com/agent-configmaps"`, that will parse and patch configurations with multiple comma-separated ConfigMaps to be mounted.

First-time contributor, and I'm not a gopher by any means. Please let me know what I can do to make this merge-ready.

Resolves #102

### DIFF
--- a/agent-inject/agent/container_volume.go
+++ b/agent-inject/agent/container_volume.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/vault/sdk/helper/strutil"
 	corev1 "k8s.io/api/core/v1"
@@ -106,6 +107,27 @@ func (a *Agent) ContainerConfigMapVolume() corev1.Volume {
 			},
 		},
 	}
+}
+
+// ContainerConfigMapVolume returns a list of volumes to mount
+// a comma-separated list of config maps if the user supplied any.
+func (a *Agent) ContainerConfigMapVolumes() []corev1.Volume {
+	var configMapNames = strings.Split(a.ConfigMapNames, ",")
+	var volumes []corev1.Volume
+	for _, configMapName := range configMapNames {
+		volumes = append(volumes,
+			corev1.Volume{
+				Name: configVolumeName,
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: configMapName,
+						},
+					},
+				},
+			})
+	}
+	return volumes
 }
 
 // ContainerExtraSecretVolume returns a volume to mount a Kube secret
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -111,6 +111,10 @@ type Agent struct {
 	// container(s).
 	ConfigMapName string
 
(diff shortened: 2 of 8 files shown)


agent-inject/agent/container_volume.go
```diff
@@ -108,6 +109,27 @@ func (a *Agent) ContainerConfigMapVolume() corev1.Volume {
 	}
 }
 
+// ContainerConfigMapVolume returns a list of volumes to mount
+// a comma-separated list of config maps if the user supplied any.
+func (a *Agent) ContainerConfigMapVolumes() []corev1.Volume {
+	var configMapNames = strings.Split(a.ConfigMapNames, ",")
```
jrcasso, line 115: Should spaces be permitted? Can trim elements of the resultant slice of strings

jasonodonnell:
Hi @jrcasso, thanks for the contribution!

Yes, this is actively maintained. Most changes in the change log are community contributions. Our team is stretched very thin however so sorry for the delay. I have some time to review this week.

jasonodonnell (commented):
Hi @jrcasso, thanks for the contribution!

I wanted to review this at a higher level to understand the feature better. The existing annotation `vault.hashicorp.com/agent-configmap` has a requirement of providing two config files (`config.hcl` and `config-init.hcl`) so we can configure the agents separately. In my opinion this seems like a reasonable requirement.

Where the existing annotation falls short is with the template files you may want to include. There could be a lot of different templates that are shared across pods, so in this case you may want to mount a set of templates.

If supporting multiple template configmaps is desirable, this PR should be scoped specifically for templates. For example, `vault.hashicorp.com/agent-configmaps` is far too similar to the existing annotation `vault.hashicorp.com/agent-configmap`.

From a user perspective something like `vault.hashicorp.com/agent-template-configmaps` seems more appropriate. We can then mount those configs to say `/vault/templates/<name of configmap>`.

Happy to discuss more.

jasonodonnell:
If we move forward with scoping this more to templates, I think `vault.hashicorp.com/agent-inject-template-file` annotation needs a bit of reworking. Currently this annotation expects templates to be mounted via `vault.hashicorp.com/agent-extra-secret`, which expects to find the files in `/vault/custom`. I think we could support custom paths but also not break backwards compatibility.

jrcasso:
Closing as stale, sorry @jasonodonnell; I just don't have the time. Thanks for considering it.

PR #262, opened 2021-06-01
Add a name to the service port

LS80:
The prometheus operator requires a named service port. This also matches the service created by the vault [helm chart](https://github.com/hashicorp/vault-helm/blob/3afcb463f8672e55c5632312228a02422eeb7cc9/templates/injector-service.yaml#L14).


LS80:
@tvoran Could you review this tiny change?

swenson (approved):
LGTM

Thanks!

PR #265, opened 2021-06-10
Update and remove deps

tvoran:
Updates yaml.v2 and golang.org/x/crypto. Removes the dependency
on consul by adding a simple version of consul's flags.Usage()
for printing command options for agent-inject.


calvn:
Does `go mod tidy` remove any of the indirect depenencies?

tvoran:
@calvn It doesn't seem to for me.

jasonodonnell (approved):

PR #266, opened 2021-06-16
Update dev dockerfile to v0.10.2

tvoran:
Post-release cleanup.


jasonodonnell (commented):
LGTM, but while in here can you also bump the UBI image to 8.4? It's now required for certification on RH registry.

PR #267, opened 2021-06-25
agent-inject: add exit_on_retry_failure flag and annotation

calvn:
This PR adds support for setting Vault Agent's new `template_config.exit_on_retry_failure` introduced in https://github.com/hashicorp/vault/pull/11775. 

A new flag and environment variable, `AGENT_INJECT_TEMPLATE_CONFIG_EXIT_ON_RETRY_FAILURE`, is added to allow this value to be toggled globally. There's also a new annotation, `agent-template-config-exit-on-retry-failure`, on agent-inject to allow for pods to override this value. The default value for this is `true` since we don't want pods being stuck during deployments if any of Agent's templates fail to render.

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -227,10 +227,10 @@ const (
 	// sidecar containers. Ignores any Kubernetes service account token mounts.
 	AnnotationAgentCopyVolumeMounts = "vault.hashicorp.com/agent-copy-volume-mounts"
 
-	// AnnotationAgentTemplateConfigExitOnRetryFailure configure whether agent
+	// AnnotationTemplateConfigExitOnRetryFailure configures whether agent
 	// will exit on template render failures once it has exhausted all its retry
 	// attempts. Defaults to true.
-	AnnotationAgentTemplateConfigExitOnRetryFailure = "vault.hashicorp.com/agent-template-config-exit-on-retry-failure"
+	AnnotationTemplateConfigExitOnRetryFailure = "vault.hashicorp.com/template-config-exit-on-retry-failure"
 )
 
 type AgentConfig struct {
@@ -399,8 +399,8 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationAgentInjectDefaultTemplate] = cfg.DefaultTemplate
 	}
 
-	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentTemplateConfigExitOnRetryFailure]; !ok {
-		pod.ObjectMeta.Annotations[AnnotationAgentTemplateConfigExitOnRetryFailure] = strconv.FormatBool(cfg.ExitOnRetryFailure)
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigExitOnRetryFailure]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationTemplateConfigExitOnRetryFailure] = strconv.FormatBool(cfg.ExitOnRetryFailure)
 	}
 
 	return nil
@@ -592,9 +592,9 @@ func (a *Agent) cacheEnable() (bool, error) {
 }
 
 func (a *Agent) templateConfigExitOnRetryFailure() (bool, error) {
-	raw, ok := a.Annotations[AnnotationAgentTemplateConfigExitOnRetryFailure]
+	raw, ok := a.Annotations[AnnotationTemplateConfigExitOnRetryFailure]
 	if !ok {
-		return DefaultAgentTemplateConfigExitOnRetryFailure, nil
+		return DefaultTemplateConfigExitOnRetryFailure, nil
 	}
 
 	return strconv.ParseBool(raw)
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -226,6 +226,11 @@ const (
 	// in the Pod whose volume mounts should be copied onto the Vault Agent init and
 	// sidecar containers. Ignores any Kubernetes service account token mounts.
 	AnnotationAgentCopyVolumeMounts = "vault.hashicorp.com/agent-copy-volume-mounts"
+
+	// AnnotationAgentTemplateConfigExitOnRetryFailure configure whether agent
+	// will exit on template render failures once it has exhausted all its retry
+	// attempts. Defaults to true.
+	AnnotationAgentTemplateConfigExitOnRetryFailure = "vault.hashicorp.com/agent-template-config-exit-on-retry-failure"
 )
 
 type AgentConfig struct {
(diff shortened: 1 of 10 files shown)


agent-inject/agent/annotations.go
```diff
@@ -226,6 +226,11 @@ const (
 	// in the Pod whose volume mounts should be copied onto the Vault Agent init and
 	// sidecar containers. Ignores any Kubernetes service account token mounts.
 	AnnotationAgentCopyVolumeMounts = "vault.hashicorp.com/agent-copy-volume-mounts"
+
+	// AnnotationAgentTemplateConfigExitOnRetryFailure configure whether agent
+	// will exit on template render failures once it has exhausted all its retry
+	// attempts. Defaults to true.
+	AnnotationAgentTemplateConfigExitOnRetryFailure = "vault.hashicorp.com/agent-template-config-exit-on-retry-failure"
```
jasonodonnell, line 233: I think we can drop `config` here. This is a really long annotation and annotations can't be larger than 63 characters (this is exactly 63 😄 ): https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

agent-inject/agent/annotations_test.go
```diff
@@ -32,6 +32,7 @@ func basicAgentConfig() AgentConfig {
 		ResourceRequestMem: DefaultResourceRequestMem,
 		ResourceLimitCPU:   DefaultResourceLimitCPU,
 		ResourceLimitMem:   DefaultResourceLimitMem,
+		ExitOnRetryFailure: true,
```
jasonodonnell, line 35: Could this instead use `DefaultAgentTemplateConfigExitOnRetryFailure`?

agent-inject/agent/container_sidecar_test.go
```diff
@@ -41,9 +41,22 @@ func TestContainerSidecarVolume(t *testing.T) {
...
+		UserID:             "1000",
+		GroupID:            "100",
+		SameID:             DefaultAgentRunAsSameUser,
+		SetSecurityContext: DefaultAgentSetSecurityContext,
+		DefaultTemplate:    "map",
+		ResourceRequestCPU: DefaultResourceRequestCPU,
+		ResourceRequestMem: DefaultResourceRequestMem,
+		ResourceLimitCPU:   DefaultResourceLimitCPU,
+		ResourceLimitMem:   DefaultResourceLimitMem,
+		ExitOnRetryFailure: true,
```
jasonodonnell, line 59: Could this instead use `DefaultAgentTemplateConfigExitOnRetryFailure`?

agent-inject/agent/container_sidecar_test.go
```diff
@@ -122,10 +135,26 @@ func TestContainerSidecarVolumeWithIRSA(t *testing.T) {
...
```
jasonodonnell, line 154: Same comment about reusing default.

agent-inject/agent/container_sidecar_test.go
```diff
@@ -184,9 +213,22 @@ func TestContainerSidecar(t *testing.T) {
...
+		GroupID:            "100",
+		SameID:             DefaultAgentRunAsSameUser,
+		SetSecurityContext: DefaultAgentSetSecurityContext,
+		ProxyAddress:       "https://proxy:3128",
+		DefaultTemplate:    "map",
+		ResourceRequestCPU: DefaultResourceRequestCPU,
+		ResourceRequestMem: DefaultResourceRequestMem,
+		ResourceLimitCPU:   DefaultResourceLimitCPU,
+		ResourceLimitMem:   DefaultResourceLimitMem,
+		ExitOnRetryFailure: true,
```
jasonodonnell, line 231: Same comment about reusing default.

agent-inject/agent/container_sidecar_test.go
```diff
@@ -303,9 +345,22 @@ func TestContainerSidecarRevokeHook(t *testing.T) {
...
+				UserID:             "1000",
+				GroupID:            "100",
+				SameID:             DefaultAgentRunAsSameUser,
+				SetSecurityContext: DefaultAgentSetSecurityContext,
+				DefaultTemplate:    "map",
+				ResourceRequestCPU: DefaultResourceRequestCPU,
+				ResourceRequestMem: DefaultResourceRequestMem,
+				ResourceLimitCPU:   DefaultResourceLimitCPU,
+				ResourceLimitMem:   DefaultResourceLimitMem,
+				ExitOnRetryFailure: true,
```
jasonodonnell, line 363: Same comment about reusing default.

agent-inject/agent/container_sidecar_test.go
```diff
@@ -358,9 +413,22 @@ func TestContainerSidecarConfigMap(t *testing.T) {
...
+		UserID:             "1000",
+		GroupID:            "100",
+		SameID:             DefaultAgentRunAsSameUser,
+		SetSecurityContext: DefaultAgentSetSecurityContext,
+		DefaultTemplate:    "map",
+		ResourceRequestCPU: DefaultResourceRequestCPU,
+		ResourceRequestMem: DefaultResourceRequestMem,
+		ResourceLimitCPU:   DefaultResourceLimitCPU,
+		ResourceLimitMem:   DefaultResourceLimitMem,
+		ExitOnRetryFailure: true,
```
jasonodonnell, line 431: Same comment about reusing default.

agent-inject/agent/container_sidecar_test.go
```diff
@@ -1060,9 +1128,22 @@ func TestContainerCache(t *testing.T) {
...
+				UserID:             "1000",
+				GroupID:            "100",
+				SameID:             DefaultAgentRunAsSameUser,
+				SetSecurityContext: DefaultAgentSetSecurityContext,
+				DefaultTemplate:    "map",
+				ResourceRequestCPU: DefaultResourceRequestCPU,
+				ResourceRequestMem: DefaultResourceRequestMem,
+				ResourceLimitCPU:   DefaultResourceLimitCPU,
+				ResourceLimitMem:   DefaultResourceLimitMem,
+				ExitOnRetryFailure: true,
```
jasonodonnell, line 1146: Same comment about reusing default.

agent-inject/agent/annotations.go
calvn: I was a bit worried that this might confuse users into thinking that this is a value on Agent's `template` stanza, but if this is a hard limit then this makes the most sense to do.

tvoran: Maybe we could drop the `agent-` prefix? I feel like that's kind of implied here.
```suggestion
	AnnotationAgentTemplateConfigExitOnRetryFailure = "vault.hashicorp.com/template-config-exit-on-retry-failure"
```

calvn: Yep, went ahead and did that in 649a051 (and also shortened the variable name).

tvoran (approved):
This tested good for me 👍 

Only thing I'd suggest is adding a test for the generated config in https://github.com/hashicorp/vault-k8s/blob/87fd9ba1348ef1a6d4d649007885cba3319d5c75/agent-inject/agent/config_test.go

calvn:
@tvoran added config tests in 10031a9, thanks for pointing that out!

PR #271, opened 2021-07-09
Moving leader election inside vault-k8s

tvoran:
This uses operator-sdk's [Become() function](https://github.com/operator-framework/operator-lib/blob/v0.4.x/leader/leader.go#L88) to determine leadership for certificate generation, instead of relying on a leader-elector container as in #198.

The associated helm chart changes are in https://github.com/hashicorp/vault-helm/pull/568

Depends on #273


leader/leader.go
```diff
@@ -1,59 +1,47 @@
...
+	go func() {
+		// This function blocks until this replica becomes the "leader", which
+		// means it creates a ConfigMap with an ownerref. Another replica can
+		// become the leader when the current leader replica stops running, and
+		// the Kubernetes garbage collector deletes the vault-k8s-leader
+		// ConfigMap.
+		err := operator_leader.Become(ctx, "vault-k8s-leader")
+		if err != nil {
+			logger.Error("trouble becoming leader:", "error", err)
+			return
```
tomhjp, line 35: Given that we don't restart the `Become` attempt, could we end up in a situation where none of the replicas are trying to become the leader, if they all error here? Seems like maybe we should have this error feed back into some sort of retry logic to guard against that.

tomhjp (commented):
Generally LGTM, though one question on resilience to errors.

leader/leader.go
```diff
@@ -1,59 +1,47 @@
...
+	le := &LeaderForLife{}
+	le.isLeader.Store(false)
+
+	go func() {
+		// This function blocks until this replica becomes the "leader", which
+		// means it creates a ConfigMap with an ownerref. Another replica can
+		// become the leader when the current leader replica stops running, and
+		// the Kubernetes garbage collector deletes the vault-k8s-leader
+		// ConfigMap.
+		err := operator_leader.Become(ctx, "vault-k8s-leader")
```
calvn, line 32: If a node calls this with a leader already elected, does this return an error, nil, or blocks until context cancelation?

deploy/injector-deployment.yaml
```diff
@@ -95,7 +67,7 @@ spec:
               port: 8080
               scheme: HTTPS
             failureThreshold: 2
-            initialDelaySeconds: 1
```
calvn, line 98: Does this change on initial delay for readiness and liveness mean that leader election is a bit longer to establish through this method, or is it to make it less flaky in general?

deploy/injector-rbac.yaml
```diff
@@ -48,13 +48,18 @@ metadata:
...
     - "create"
     - "get"
     - "watch"
     - "list"
     - "update"
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs:
+    - "get"
+    - "patch"
```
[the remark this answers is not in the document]
tvoran: It's part of the `Become()` call; it looks like there's some checking and cleanup around situations where an old leader pod was evicted or the underlying node is NotReady: https://github.com/operator-framework/operator-lib/blob/v0.4.1/leader/leader.go#L166-L189

leader/leader.go
tvoran: It looks like it blocks forever until it can achieve leadership. The context passed in is used for k8s API calls, but I don't see it being checked for cancellation in Become(): https://github.com/operator-framework/operator-lib/blob/v0.4.1/leader/leader.go#L88

deploy/injector-deployment.yaml
tvoran: Yeah, I think it was just taking a little longer than a second while I was testing it locally to achieve leadership and generate the certs so that the liveness probe would pass.

benashz (approved):
LGTM!

leader/leader.go
```diff
@@ -1,59 +1,88 @@
...
+		if err != nil {
+			logger.Error("trouble becoming leader:", "error", err)
+			return
+		}
+		le.isLeader.Store(true)
+		if err := setLeaderLabel(ctx, clientset, true); err != nil {
```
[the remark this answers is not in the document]
tvoran: Hmm, I removed the leader labeling from this PR in https://github.com/hashicorp/vault-k8s/pull/271/commits/6d837ced87f11f8cd4f410c526d33553c3632861

leader/leader.go
tvoran: @benashz Hmm, I added retries on error in https://github.com/hashicorp/vault-k8s/pull/271/commits/75258f728fb7e49cff1322925b15ceae4110af7a, I wonder if you're looking at an old diff?

Exiting right away is an interesting idea too.

tvoran: The more I think about this, the more I think it would be better for this to retry a few times, then if Become() is still failing, throw a signal or something that would do a system exit non-zero. Otherwise folks would need to be watching the logs to know there was a problem with the replica.

tvoran: @benashz @tomhjp added retries then exit non-zero in b73527f.

leader/leader.go
```diff
@@ -1,59 +1,88 @@
...
-	if err != nil {
-		return false, err
+func (le *LeaderForLife) IsLeader() (bool, error) {
+	leaderStatus := le.isLeader.Load().(bool)
+	return leaderStatus, nil
+}
+
+func setLeaderLabel(ctx context.Context, clientset kubernetes.Interface, label bool) error {
+	namespace := os.Getenv("NAMESPACE")
+	if namespace == "" {
```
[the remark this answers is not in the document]
tvoran: It looks like this comment was on an old commit, but as for checking for POD_NAME, Become() will return an error if POD_NAME isn't set. And since we're handling that error with a finite amount of retries, it'll eventually case the injector to exit non-zero.

leader/leader.go
tomhjp: I like that line of thinking 👍

subcommand/injector/command.go
```diff
@@ -218,11 +221,12 @@ func (c *Command) Run(args []string) int {
 	}()
 	go func() {
 		select {
+		case <-exitOnError:
+			logger.Error("shutting down due to errors")
```
tomhjp, line 225: As per the comment above, I think this log message should include the error that triggered the shutdown.

tomhjp (approved):
LGTM - I think the failure mode flow looks good, just one suggestion on the logging.

subcommand/injector/command.go
```diff
@@ -296,18 +307,17 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
 		isLeader := true
 		if c.flagUseLeaderElector {
 			// Only the leader should do the caBundle patching in k8s API
 			var err error
-			le := leader.New()
-			isLeader, err = le.IsLeader()
+			isLeader, err = leaderElector.IsLeader()
 			if err != nil {
-				log.Error(fmt.Sprintf("error checking leader: %s", err))
+				log.Warn(fmt.Sprintf("Could not check leader: %s. Trying again...", err))
```
benashz, line 320: I noticed that some log lines are capitalized, and in other cases not. I guess we should probably be consistent either way.

subcommand/injector/command.go
benashz: It looks like on `exitOnError` does not result in a non-zero status being returned to the caller. I presumes that would make its way to `os.Exit()`.

tvoran: > It looks like on exitOnError does not result in a non-zero status being returned to the caller. I presumed that it would make its way to os.Exit().

The flow here is:
* `c.shutdownHandler()` calls `server.Shutdown()`
* `server.Shutdown()` causes `server.ListenAndServeTLS()` to return an error, triggering a `return 1` for this function:
https://github.com/hashicorp/vault-k8s/blob/b73527fa89b29ec705daf58e49b03bd81a524af7/subcommand/injector/command.go#L237-L240

And that makes its way back to os.Exit in main.go:
https://github.com/hashicorp/vault-k8s/blob/b73527fa89b29ec705daf58e49b03bd81a524af7/main.go#L17-L21

subcommand/injector/command.go
tvoran: Sure, I think generally log messages should start with a capitalized word, and errors themselves should start with lowercase.

leader/leader.go
```diff
@@ -1,59 +1,65 @@
...
+				logger.Error("trouble becoming leader", "error", err)
+				return err
+			}
+			return nil
+		}, bo)
+
+		if err != nil {
+			// Signal the caller to shutdown the injector server, since Become()
+			// failed all the retries
+			exitOnError <- true
```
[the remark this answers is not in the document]
tvoran: Great idea! Changed in a45b35e.

subcommand/injector/command.go
tvoran: > As per the comment above, I think this log message should include the error that triggered the shutdown.

Added the error to the log message in a45b35e

benashz: Thanks for the explanation. I guess we are relying on the fact that `server.Shutdown()` is called before `server.ListenAndServeTLS()`?

subcommand/injector/command.go
benashz: Totally agree 👍

benashz (commented):
Looking good. Only one request to add comment for `exitOnError`

PR #273, opened 2021-07-26
Update client go to latest

tvoran:
Update client-go to v0.21.3. Adds support for switching between v1 and v1beta1 Admission APIs based on availability in the cluster. Since the v1 and v1beta1 AdmissionReview, AdmissionResponse, and AdmissionRequest are the same, this just bumps those to v1. Checks at the start of `certWatcher()` if AdmissionregistrationV1 exists, and falls back to v1beta1 if it doesn't.

This also is necessary for running on k8s 1.22: https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes

Tested with our injector acceptance tests on k8s 1.14 and 1.21.

(Split out from #271 for clarity).


subcommand/injector/command.go
```diff
@@ -291,15 +306,28 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 		if isLeader && c.flagAutoName != "" && len(bundle.CACert) > 0 {
 			// The CA Bundle value must be base64 encoded
 			value := base64.StdEncoding.EncodeToString(bundle.CACert)
+			payload := fmt.Sprintf(
```
benashz, line 309: Wonder if we should make `payload` a byte slice to avoid the duplicated conversion below.

subcommand/injector/command.go
```diff
@@ -291,15 +306,28 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
+					MutatingWebhookConfigurations().
+					Patch(ctx, c.flagAutoName, types.JSONPatchType, []byte(payload),
+						metav1.PatchOptions{})
+			// AdmissionregistrationV1beta1 is present in k8s 1.9 - 1.19
+			case "v1beta1":
+				_, err = clientset.AdmissionregistrationV1beta1().
+					MutatingWebhookConfigurations().
+					Patch(ctx, c.flagAutoName, types.JSONPatchType, []byte(payload),
+						metav1.PatchOptions{})
+			}
```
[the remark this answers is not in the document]
tvoran: Sure, added in 91474cf92a00037fb2e66cbf76359412819ee412.

subcommand/injector/command.go
tvoran: Done in 91474cf92a00037fb2e66cbf76359412819ee412.

agent-inject/handler.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/hashicorp/vault-k8s/agent-inject/agent"
 	"github.com/hashicorp/vault/sdk/helper/strutil"
 	"github.com/mattbaird/jsonpatch"
-	"k8s.io/api/admission/v1beta1"
+	v1 "k8s.io/api/admission/v1"
```
tomhjp, line 14: nit: The alias seems unnecessary as-is, but should this be `admissionv1` to follow the example of other similar packages?

tomhjp (approved):
LGTM

benashz (approved):
LGTM!

agent-inject/handler.go
tvoran: Yep, that does look odd. Fixed in fe858b700c4f856183c027c91a8cdaca8b798f3e.

PR #276, opened 2021-08-04
Feat: static_secret_render_interval

burdandrei:
Adds option to configure template config static_secret_render_interval

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -4,33 +4,33 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
-	"strings"
-
 	"github.com/mattbaird/jsonpatch"
 	corev1 "k8s.io/api/core/v1"
+	"strconv"
+	"strings"
 )
 
 // TODO swap out 'github.com/mattbaird/jsonpatch' for 'github.com/evanphx/json-patch'
 
 const (
-	DefaultVaultImage                       = "hashicorp/vault:1.8.0"
-	DefaultVaultAuthType                    = "kubernetes"
-	DefaultVaultAuthPath                    = "auth/kubernetes"
-	DefaultAgentRunAsUser                   = 100
-	DefaultAgentRunAsGroup                  = 1000
-	DefaultAgentRunAsSameUser               = false
-	DefaultAgentAllowPrivilegeEscalation    = false
-	DefaultAgentDropCapabilities            = "ALL"
-	DefaultAgentSetSecurityContext          = true
-	DefaultAgentReadOnlyRoot                = true
-	DefaultAgentCacheEnable                 = "false"
-	DefaultAgentCacheUseAutoAuthToken       = "true"
-	DefaultAgentCacheListenerPort           = "8200"
-	DefaultAgentCacheExitOnErr              = false
-	DefaultAgentUseLeaderElector            = false
-	DefaultAgentInjectToken                 = false
-	DefaultTemplateConfigExitOnRetryFailure = true
+	DefaultVaultImage                               = "hashicorp/vault:1.8.0"
+	DefaultVaultAuthType                            = "kubernetes"
+	DefaultVaultAuthPath                            = "auth/kubernetes"
+	DefaultAgentRunAsUser                           = 100
+	DefaultAgentRunAsGroup                          = 1000
+	DefaultAgentRunAsSameUser                       = false
+	DefaultAgentAllowPrivilegeEscalation            = false
+	DefaultAgentDropCapabilities                    = "ALL"
+	DefaultAgentSetSecurityContext                  = true
+	DefaultAgentReadOnlyRoot                        = true
+	DefaultAgentCacheEnable                         = "false"
+	DefaultAgentCacheUseAutoAuthToken               = "true"
+	DefaultAgentCacheListenerPort                   = "8200"
+	DefaultAgentCacheExitOnErr                      = false
+	DefaultAgentUseLeaderElector                    = false
+	DefaultAgentInjectToken                         = false
+	DefaultTemplateConfigExitOnRetryFailure         = true
+	DefaultTemplateConfigStaticSecretRenderInterval = "5m"
 )
 
 // Agent is the top level structure holding all the
@@ -270,6 +270,10 @@ type VaultAgentTemplateConfig struct {
 	// ExitOnRetryFailure configures whether agent should exit after failing
 	// all its retry attempts when rendering templates
 	ExitOnRetryFailure bool
+
+	// StaticSecretRenderInterval If specified, configures how often
+	// Vault Agent Template should render non-leased secrets such as KV v2
+	StaticSecretRenderInterval string
 }
 
 // New creates a new instance of Agent by parsing all the Kubernetes annotations.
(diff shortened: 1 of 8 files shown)


agent-inject/agent/annotations.go
```diff
@@ -231,26 +231,32 @@ const (
 	// will exit on template render failures once it has exhausted all its retry
 	// attempts. Defaults to true.
 	AnnotationTemplateConfigExitOnRetryFailure = "vault.hashicorp.com/template-config-exit-on-retry-failure"
+
+	// AnnotationTemplateConfigStaticSecretRenderInterval
+	// If specified, configures how often Vault Agent Template should render non-leased secrets such as KV v2.
+	// Defaults to 5 minutes.
+	AnnotationTemplateConfigStaticSecretRenderInterval = "vault.hashicorp.com/template-config-static-secret-render-interval"
```
tvoran, line 238: I think this is over the 63-character limit for annotation names: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

So we'll need to pick something shorter. Perhaps `template-static-secret-render-interval`?

burdandrei: Thanks for the hint, @tvoran, working on the fix

agent-inject/agent/annotations.go
```diff
@@ -402,6 +408,9 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigExitOnRetryFailure]; !ok {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigExitOnRetryFailure] = strconv.FormatBool(cfg.ExitOnRetryFailure)
 	}
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval] = DefaultTemplateConfigStaticSecretRenderInterval
```
tvoran, line 412: This will need to be set to cfg.StaticSecretRenderInterval if we want to set it via a command line flag/env variable:
```suggestion
		pod.ObjectMeta.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval] = cfg.StaticSecretRenderInterval
```

agent-inject/agent/config.go
```diff
@@ -103,7 +103,8 @@ type CachePersist struct {
 
 // TemplateConfig defines the configuration for template_config in Vault Agent
 type TemplateConfig struct {
-	ExitOnRetryFailure bool `json:"exit_on_retry_failure"`
+	ExitOnRetryFailure         bool   `json:"exit_on_retry_failure"`
+	StaticSecretRenderInterval string `json:"static_secret_render_interval"`
```
tvoran, line 107: If we're not setting a default as suggested above, then adding `omitempty` here will prevent that option from being rendered unless it was set by the user:
```suggestion
	StaticSecretRenderInterval string `json:"static_secret_render_interval,omitempty"`
```

agent-inject/agent/agent.go
```diff
@@ -4,33 +4,33 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
-	"strings"
-
 	"github.com/mattbaird/jsonpatch"
 	corev1 "k8s.io/api/core/v1"
+	"strconv"
+	"strings"
```
tvoran, line 10: Hmm, generally we keep 3rd party packages separate from built-in:
```suggestion
	"strconv"
	"strings"

	"github.com/mattbaird/jsonpatch"
	corev1 "k8s.io/api/core/v1"
```

agent-inject/agent/agent.go
```diff
@@ -4,33 +4,33 @@ import (
...
+	DefaultAgentSetSecurityContext                  = true
+	DefaultAgentReadOnlyRoot                        = true
+	DefaultAgentCacheEnable                         = "false"
+	DefaultAgentCacheUseAutoAuthToken               = "true"
+	DefaultAgentCacheListenerPort                   = "8200"
+	DefaultAgentCacheExitOnErr                      = false
+	DefaultAgentUseLeaderElector                    = false
+	DefaultAgentInjectToken                         = false
+	DefaultTemplateConfigExitOnRetryFailure         = true
+	DefaultTemplateConfigStaticSecretRenderInterval = "5m"
```
tvoran, line 33: If the user hasn't specified a render interval, I'm thinking we should just not set it and let the default on the vault agent side be the default.
```suggestion
```

agent-inject/agent/agent.go
```diff
@@ -412,9 +416,14 @@ func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, erro
 	if err != nil {
 		return nil, err
 	}
+	staticSecretRenderInterval, err := agent.templateConfigStaticSecretRenderInterval()
+	if err != nil {
+		return nil, err
+	}
```
tvoran, line 422: If we're not setting a default as suggested above, this can be set directly in the VaultAgentTemplateConfig struct.
```suggestion
```

agent-inject/agent/agent.go
```diff
@@ -412,9 +416,14 @@ func New(pod *corev1.Pod, patches []*jsonpatch.JsonPatchOperation) (*Agent, erro
...
 
 	agent.VaultAgentTemplateConfig = VaultAgentTemplateConfig{
-		ExitOnRetryFailure: exitOnRetryFailure,
+		ExitOnRetryFailure:         exitOnRetryFailure,
+		StaticSecretRenderInterval: staticSecretRenderInterval,
```
tvoran, line 426:
```suggestion
		StaticSecretRenderInterval: pod.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval],
```

agent-inject/agent/annotations.go
```diff
@@ -600,6 +609,15 @@ func (a *Agent) templateConfigExitOnRetryFailure() (bool, error) {
...
 
+func (a *Agent) templateConfigStaticSecretRenderInterval() (string, error) {
+	raw, ok := a.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval]
+	if !ok {
+		return DefaultTemplateConfigStaticSecretRenderInterval, nil
+	}
+
+	return raw, nil
+}
+
```
tvoran, line 620: This function is no longer needed if we're not setting a default as suggested above.
```suggestion
```

subcommand/injector/flags.go
```diff
@@ -107,6 +111,8 @@ func (c *Command) init() {
 		`Supported log formats: "standard", "json".`)
 	c.flagSet.BoolVar(&c.flagExitOnRetryFailure, "template-config-exit-on-retry-failure", agent.DefaultTemplateConfigExitOnRetryFailure,
 		fmt.Sprintf("Value for Agent's template_config.exit_on_retry_failure. Defaults to %t.", agent.DefaultTemplateConfigExitOnRetryFailure))
+	c.flagSet.StringVar(&c.flagStaticSecretRenderInterval, "template-static-secret-render-interval", agent.DefaultTemplateConfigStaticSecretRenderInterval,
+		fmt.Sprintf("Value for Agent's template_config.exit_on_retry_failure. Defaults to %s.", agent.DefaultTemplateConfigStaticSecretRenderInterval))
```
tvoran, line 115:
```suggestion
	c.flagSet.StringVar(&c.flagStaticSecretRenderInterval, "template-static-secret-render-interval", "",
		"Value for Agent's template_config.exit_on_retry_failure.")
```

tvoran (commented):
Just some suggestions, mostly around the idea that we don't need to set a default for this config parameter, since vault agent has one already if the option is omitted from the config.

agent-inject/agent/agent.go
burdandrei: yeah, sorry, Goland's fault =)

burdandrei:
Thanks for suggestions @tvoran!
Everything committed and tests passed.
Also, PRs include only CLA check right now.
Maybe we should add unit test run as a check as well?

tvoran (approved):
Thanks!

tvoran:
@burdandrei Yeah, the unit tests don't run on PRs from forks. We may want to revisit that in the future.

PR #280, opened 2021-08-10
Continuously retry updating the cert secret

tvoran:
Uses retry with exponential backoff to retry updating the cert secret
if there's an error. Sync's with caller via a stop channel, caller
waits on stop channel before starting a new update thread.


helper/cert/source_gen.go
```diff
@@ -143,9 +150,19 @@ func (s *GenSource) Certificate(ctx context.Context, last *Bundle) (Bundle, erro
 		result.Key = []byte(key)
```
tomhjp, line 150: Not new in this PR, but this block should probably de-indent out of the `if err == nil` braces, and instead return early `if err != nil`.

helper/cert/source_gen.go
```diff
@@ -143,9 +150,19 @@ func (s *GenSource) Certificate(ctx context.Context, last *Bundle) (Bundle, erro
...
+				s.UpdateCancel()
+				<-s.StopUpdate
+			}
+			// Start a new update goroutine
+			var updateCtx context.Context
+			updateCtx, s.UpdateCancel = context.WithCancel(ctx)
+			if s.StopUpdate == nil {
+				s.StopUpdate = make(chan struct{})
 			}
+			go s.retryUpdateSecret(updateCtx, result)
```
tomhjp, line 165: AFAIU, this is run in a goroutine to allow the function to return the certificate before the k8s secret is updated, in case that takes a large number of retries? And that seems to add a fair bit of complexity and statefulness to the control flow, because now you have to synchronise across multiple calls to `Certificate`.

I definitely could be missing some concerns or requirements, but is that complexity necessary? It seems like it would be ok to synchronously update the k8s secret before returning, and then you could remove a bunch of the syncing required.

tomhjp (commented):
LGTM, mainly just one question about whether we can simplify a little.

helper/cert/source_gen.go
```diff
@@ -182,6 +199,33 @@ func (s *GenSource) checkLeader(ctx context.Context, changed chan<- bool) {
...
+	}()
+
+	// New exponential backoff with unlimited retries
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = time.Second * 30
+	bo.MaxElapsedTime = 0
+	ticker := backoff.NewTicker(bo)
+
+	for {
+		select {
```
tomhjp, line 214: What happens if there is a leadership change during this retry loop? Should we also cancel then?

helper/cert/source_gen.go
```diff
@@ -61,6 +62,12 @@ type GenSource struct {
 	SecretsCache  informerv1.SecretInformer
 	LeaderElector *leader.LeaderElector
 
+	// UpdateCancel and StopUpdate are used to control and sync the goroutine
+	// that continuously attempts to update the K8s Secret with the new
+	// certificate
+	UpdateCancel context.CancelFunc
+	StopUpdate   chan struct{}
+
```
calvn, line 70: You can look into [oklog/run](https://github.com/oklog/run) to do this kind of synchronization.

calvn: I misread this! It doesn't look like you're spawning multiple goroutines. It's more like handling the lifecycle for a single goroutine that can live past the `Certificate()` call.

helper/cert/source_gen.go
calvn: I agree with this sentiment, and it makes the guarantee of `result` a bit weak on whoever calls `Certificate()` since the update on the k8s is not guaranteed to have occurred (or can just flat out error) before the caller consumes the value.

helper/cert/source_gen.go
```diff
@@ -182,6 +199,33 @@ func (s *GenSource) checkLeader(ctx context.Context, changed chan<- bool) {
...
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = time.Second * 30
+	bo.MaxElapsedTime = 0
+	ticker := backoff.NewTicker(bo)
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := s.updateSecret(ctx, bundle); err != nil {
+				s.Log.Error(fmt.Sprintf("attempt to update Secret %q failed: %s", certSecretName, err))
```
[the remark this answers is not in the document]
tvoran: Seems like a good idea to me 👍

helper/cert/source_gen.go
tvoran: Good question, and yes, I think we should also cancel then. Added in 51f676243f146b93c2836e9905f09ff69e15a00b.

helper/cert/source_gen.go
tvoran: Yeah, I think those are good points, and `Certificate()` is already being called from another goroutine anyway. Changed in 51f676243f146b93c2836e9905f09ff69e15a00b, and set the max time for retrying to be 30 minutes.

helper/cert/source_gen.go
tvoran: Done in 51f676243f146b93c2836e9905f09ff69e15a00b.

helper/cert/source_gen.go
```diff
@@ -138,15 +139,14 @@ func (s *GenSource) Certificate(ctx context.Context, last *Bundle) (Bundle, erro
...
+	result.Cert = []byte(cert)
+	result.Key = []byte(key)
 
-		if s.LeaderElector != nil {
-			if err := s.updateSecret(ctx, result); err != nil {
-				return result, fmt.Errorf("failed to update Secret: %s", err)
-			}
-		}
+	if s.LeaderElector != nil {
+		s.retryUpdateSecret(ctx, leaderCh, result)
```
calvn, line 149: This sync call might affect `<-retryTicker.C` in `Notify.Run()`? For instance, if this call blocks for 6s it will be skipping the next tick (and every increment of 5 seconds that this blocks skips 1 tick). This may or may not be fine depending if there are other things within that loop that is expecting the ticker to run every 5 seconds.

helper/cert/source_gen.go
```diff
@@ -138,15 +139,14 @@ func (s *GenSource) Certificate(ctx context.Context, last *Bundle) (Bundle, erro
 
 	// Generate cert, set it on the result, and return
 	cert, key, err := s.generateCert()
-	if err == nil {
-		result.Cert = []byte(cert)
-		result.Key = []byte(key)
+	if err != nil {
+		return result, err
```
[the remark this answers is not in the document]
tomhjp: I think this is probably technically ok because the result is discarded if there's an error, but I'll let Theron confirm.

helper/cert/source_gen.go
tomhjp: This looks fine to me, the `Notify.Run()` loop only notifies upstream when there is a new result to report, so I think it should be ok to reattempt until we get something worth reporting, and we're still cancellable via the context.

helper/cert/source_gen.go
```diff
@@ -182,6 +182,32 @@ func (s *GenSource) checkLeader(ctx context.Context, changed chan<- bool) {
 	}
 }
 
+func (s *GenSource) retryUpdateSecret(ctx context.Context, leaderCh chan bool, bundle Bundle) {
```
tomhjp, line 185: I know we log errors as we retry, but it seems a bit odd to me to have finite retries and no error return value. From a user perspective, it would be nice to know that it gave up and restarted the whole process in between all the retry errors, so perhaps this function should return a terminal error on falling out of the retry loop, and then the `Notify.Run()` function will log it?

Although I would also be fine with infinite retries given we are cancellable through the context and leader elections. WDYT?

tomhjp (approved):
One more small suggestion. LGTM otherwise!

helper/cert/source_gen.go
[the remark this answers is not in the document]
tvoran: It is indeed being discarded in `notify.Run()` if there's an error. Which is good, because the next block there is comparing the new result to the old one, so it would also think an empty result is a new result. All this to say I think we might as well leave it up to the caller when returning an error, and technically `s.caCert` is valid at this point and will continue to be used on the next pass.

helper/cert/source_gen.go
tvoran: Yep, I agree that does look odd without a return value, added in https://github.com/hashicorp/vault-k8s/pull/280/commits/c37fea7f6d9c50ef4fb3643125d203b3b086d9f9. I ended up changing the max time for the retries to be 90% of the certificate's life, since that's the point when the cert will be regenerated in `Certificate()`.

helper/cert/source_gen.go
```diff
@@ -182,30 +184,38 @@ func (s *GenSource) checkLeader(ctx context.Context, changed chan<- bool) {
...
+	// by the calling function
 	bo := backoff.NewExponentialBackOff()
 	bo.MaxInterval = 30 * time.Second
-	bo.MaxElapsedTime = 30 * time.Minute
+	bo.MaxElapsedTime = time.Duration(float64(s.expiry()) * 0.90)
 	ticker := backoff.NewTicker(bo)
 	defer ticker.Stop()
-	for {
+
+	for range ticker.C {
```
tomhjp, line 197: I'm not sure how important this is, but I think this `range` loop means we may wait up to `bo.MaxInterval` before returning when a leadership change happens or the context is cancelled. I think you can fix this by switching back to the same `for`/`select` pattern as before with a small modification to the `ticker.C` case:

```go
var err error
for {
	select {
	case _, ok := <-ticker.C:
		if !ok {
			// channel closed implies timeout
			return fmt.Errorf("timed out attempting to update cert secret: %w", err)
		}
		err = s.updateSecret(...)
	case ...:
		...
	// No default case
}
```

tvoran: Ah yes! I forgot you could do that with channels. Changed in 8cded14.

tomhjp (approved):
LGTM!

PR #287, opened 2021-08-23
Keep the last CA when creating a new one

tvoran:
Keeps the last CA in the K8s API caBundle when generating a new CA. Should cover the case when follower replicas haven't received a new leaf cert from the new CA, and are still listening with a leaf cert from the previous CA.

### DIFF
--- a/helper/cert/source_gen.go
+++ b/helper/cert/source_gen.go
@@ -8,6 +8,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/sha256"
+	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
@@ -20,6 +21,8 @@ import (
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault-k8s/leader"
+	adminv1 "k8s.io/api/admissionregistration/v1"
+	adminv1beta "k8s.io/api/admissionregistration/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


helper/cert/source_gen.go
```diff
@@ -393,3 +445,41 @@ func parseCert(pemValue []byte) (*x509.Certificate, error) {
...
+			certs.Certificate = append(certs.Certificate, next.Bytes)
+		}
+	}
+	return certs
+}
+
+// appendCert returns a new CA bundle:
+// [0] last CA cert from oldCABundle
+// [1] newCACert
+func appendCert(oldCABundle, newCACert []byte) ([]byte, error) {
```
tomhjp, line 469: I think this name could be a bit better to capture that it's smarter than a simple append. Maybe `prependPreviousCA`, `prependLastCA`, or `includePreviousCA`, possibly with the argument order swapped? As of course the ordering is important too, because we always want to keep the CAs in age order.

helper/cert/source_gen.go
```diff
@@ -393,3 +445,41 @@ func parseCert(pemValue []byte) (*x509.Certificate, error) {
...
+func decodeCerts(caBundle []byte) tls.Certificate {
+	certs := tls.Certificate{}
+	remainder := caBundle
+	var next *pem.Block
+	for {
+		next, remainder = pem.Decode(remainder)
+		if next == nil {
+			break
+		}
+		if next.Type == "CERTIFICATE" {
```
tomhjp, line 459: What might we throw away in the else clause, we would never expect anything right? Perhaps worth logging a warning to surface anything unexpected?

helper/cert/source_gen.go
```diff
@@ -217,6 +235,40 @@ func (s *GenSource) getBundleFromSecret() (Bundle, error) {
 	return bundle, nil
 }
 
+func (s *GenSource) getExistingCA(ctx context.Context) []byte {
+	var caBundle []byte
```
tomhjp, line 239: nit: My personal preference is for less mutability where practical, and I think every case in this switch could just immediately return once it finds a result, rather than needing to track a variable throughout the function.

helper/cert/source_gen_test.go
```diff
@@ -237,3 +240,132 @@ func testGetHostname(t *testing.T) string {
...
+		},
+		"invalid old CAs": {
+			oldCAs:   []byte("not a cert"),
+			expected: newBundle.CACert,
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			result, err := appendCert(tc.oldCAs, newBundle.CACert)
```
tomhjp, line 281: Is it worth adding a test case where you go through an additional generation of `appendCert`? I was thinking yes because of the fact it consumes its own output (eventually), and so it would be nice to ensure by induction that it can continue forever, but after thinking about it a bit more I'm not sure if it really adds much to the coverage. Up to you if you think it has value 🙂

tomhjp (approved):
Just some nits, looks great!

Issue #290, opened 2021-08-30
Allow configurable init and sidecar container names


pbar1:
**Is your feature request related to a problem? Please describe.**
The agent injector currently injects init and sidecar containers named `vault-agent-init` and `vault-agent` respectively. This is fine and makes things clear for normal operation. However, if the pod already has containers with those names (ie - if multiple Vault injectors exist on a cluster), there will be a name collision. These names are currently hardcoded.

**Describe the solution you'd like**
`vault-k8s` should support annotations to configure the agent init and sidecar container names, as well as CLI flags and environment variables to configure global defaults for these values. The default values should be preset to the existing hardcoded values. 

**Describe alternatives you've considered**
As the current container names are set in stone, the only alternative would be to get the containers that these names collide with to change their names.

**Additional context**
N/A

dtrts:
Hi @tomhjp, 

I'm in a similar situation as @pbar1 and have looked through the issue you linked. 

It is my understanding that the annotation is to be applied to the pods which are being injected into. Is it possible to put the annotations onto the vault-agent-injector pod itself to perform the renaming before it is inserted into the target pod?

tomhjp:
@dtrts I'm not sure if I quite understand the request. You're correct that the annotation is applied to the pod being injected into. And injection is still an atomic operation, i.e. the patch is applied to the generated container spec before it gets added to the pod spec. Does that answer your question?

dtrts:
@tomhjp Thanks for the quick response. That does answer my question, although i think it means the approach detailed in #291 might not be suitable for the problem on this issue. 

If the annotation resides only on the target pod then how do we alter the name of the agent containers related to their source? 

---- 

In my situation I have two deployments of the agent in the same k8s cluster, agent-A and agent-B. These agents are configured to select mutually exclusive pods. For example pod-A-1 has a label which will trigger agent-A, pod-B-2 will trigger agent-B. 

To add clarity to my team, and run tests I was looking for a way to alter the container names so that when looking at pod-A-1 for example it was super obvious which agent had run. 

I'm looking for a way to configure the agent-container name (in the target pod) to signify which agent it was triggered by. If the annotation only works on the target-pod then there isn't an easy way to define that annotation to display what I was looking for. 

Is there some detailed JSON path I can define to differentiate the source agent? 

---- 

Hopefully i have communicated that clearly. Ultimately it would be nice if the sidecar/init containers could be configured on the Agent side of the process, instead of the Target Pod side of the process.

[...]

PR #291, opened 2021-08-30
Make agent init and sidecar container names configurable

pbar1:
This PR adds the feature mentioned in #290.

Currently, the agent init and sidecar containers that are injected are named `vault-agent-init` and `vault-agent`. Here we introduce annotations to configure them on-demand, and flags/env vars to specify them globally. Default behavior remains the same.

| Annotation                                         | Flag                      | Environment Variable                               | Default            |
|----------------------------------------------------|---------------------------|---------------------------------------|--------------------|
| `vault.hashicorp.com/agent-init-container-name`    | `-init-container-name`    | `AGENT_INJECT_INIT_CONTAINER_NAME`    | `vault-agent-init` |
| `vault.hashicorp.com/agent-sidecar-container-name` | `-sidecar-container-name` | `AGENT_INJECT_SIDECAR_CONTAINER_NAME` | `vault-agent`      |


tomhjp:
Sorry this has been left unattended for so long. In the meantime, we have added a new more generic annotation that allows users to patch the injected sidecars with an arbitrary JSON patch: https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-json-patch. The intention here is to allow an escape hatch for very detailed behaviour without users being blocked on, or overwhelmed by, a specific annotation for each feature request. The docs site includes an example of what this will look like for updating the container name: `[{"op": "replace", "path": "/name", "value": "different-name"}]`.

As such, I'm going to close this and #290. Thank you for the request and following it up with a PR, please do let me know if you don't think the above satisfies the original request.

Issue #294, opened 2021-09-04
webhook should write typed AdmissionReview response with GVK set


ravilr:
**Describe the bug**
Running the vault-injector with this webhookconfiguration https://github.com/hashicorp/vault-k8s/blob/v0.12.0/deploy/injector-mutating-webhook.yaml#L12 leads to all webhook requests failing with below error in kube-apiserver logs:

`failed calling webhook \"vault.hashicorp.com\": expected webhook response of admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind="`

by changing the order of `admissionReviewVersions: ["v1beta1", "v1"]`  fixes the above error on kube-apiserver version < v1.22.

**To Reproduce**
- Deploy vault-agent-injector v0.12.0 with webhookconfiguration having `admissionReviewVersions: ["v1", "v1beta1"]` and `failurePolicy: fail`. All pod creation requests fails with above error.

**Expected behavior**
webhook must return a well formed AdmissionReview response including apiVersion and kind (GVK). And to support multiple admissionReviewVersions configured in webhookconfiguration, webhook should look at the GVK typemeta of the incoming AdmissionReview request and send the response with Typemeta set to the same GVK as the one in the admission request. 

**Environment**
* Kubernetes version: v1.21.4
* vault-k8s version: v0.12.0

**Additional context**
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.10.0/pkg/webhook/admission/http.go#L79 is relevant.

tvoran:
Hi @ravilr, thanks for looking into this! We've done most of our testing with the templates in [vault-helm](https://github.com/hashicorp/vault-helm/blob/8932c372b75be87540f30b194eac587c85b4a3d4/templates/injector-mutating-webhook.yaml#L18-L20), which has the ordering of `admissionReviewVersions: ["v1beta1", "v1"]`, so I'm guessing that's why we haven't run into this in testing.

Would it be sufficient to just change the ordering here in the vault-k8s deployment yaml as well?

ravilr:
> Would it be sufficient to just change the ordering here in the vault-k8s deployment yaml as well?

[...]

PR #296, opened 2021-09-07
set GVK on AdmissionReview responses in webhook

ravilr:
- Fixes https://github.com/hashicorp/vault-k8s/issues/294

Specifying multiple admissionReviewVersions in webhook config with v1 as the first version causes untyped (no GVK) AdmissionReview responses to fail with an error: 

`failed calling webhook \"vault.hashicorp.com\": expected webhook response of admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind="`

This error results from the kube-apiserver checking response type for v1.AdmissionReview's but not for v1beta1's. Hence, set the AdmissionReview response object's GVK to the request's type if it exists or defaulting to v1.  with this change, either`admissionReviewVersions: ["v1beta1", "v1"]` or `admissionReviewVersions: ["v1", "v1beta1"]` in webhookconfiguration spec works correctly as expected.

- Also fix a typo in `deploy/injector-mutating-webhook.yaml` manifest.

### DIFF
--- a/agent-inject/agent/config.go
+++ b/agent-inject/agent/config.go
@@ -177,7 +177,7 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		},
 		Templates: a.newTemplateConfigs(),
 		TemplateConfig: &TemplateConfig{
-			ExitOnRetryFailure: a.VaultAgentTemplateConfig.ExitOnRetryFailure,
+			ExitOnRetryFailure:         a.VaultAgentTemplateConfig.ExitOnRetryFailure,
 			StaticSecretRenderInterval: a.VaultAgentTemplateConfig.StaticSecretRenderInterval,
 		},
 	}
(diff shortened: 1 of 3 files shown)


agent-inject/agent/config.go
```diff
@@ -177,7 +177,7 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		},
 		Templates: a.newTemplateConfigs(),
 		TemplateConfig: &TemplateConfig{
-			ExitOnRetryFailure: a.VaultAgentTemplateConfig.ExitOnRetryFailure,
+			ExitOnRetryFailure:         a.VaultAgentTemplateConfig.ExitOnRetryFailure,
```
ravilr, line 180: `make fmt`/gofmt change.

ravilr:
@tvoran PTAL as this is updating https://github.com/hashicorp/vault-k8s/pull/273..

tvoran (approved):
Thanks! This tested great for me, much appreciated.

PR #297, opened 2021-09-10
Change the leader elector log message

tvoran:
So it's easier to tell when the new internal version is being used, vs the old container version.

So the previous version of vault-k8s logs a message like:

```
Using leader elector logic
```

And with this PR, the next version of vault-k8s will log a message like:

```
Using internal leader elector logic for webhook certificate management
```


benashz (approved):
LGTM!

PR #302, opened 2021-10-19
TLS server options

tvoran:
Adds options for setting the TLS minimum version and supported cipher suites. Each can be set via command-line flag or environment variable.

| Environment variable | command-line flag | default | allowed values |
| --------------------  | ---------------- | ------- | ------------ |
| AGENT_INJECT_TLS_MIN_VERSION | -tls-min-version | tls12 | tls10 tls11 tls12 tls13 |
| AGENT_INJECT_TLS_CIPHER_SUITES | -tls-cipher-suites | "" | comma-separated list of cipher suites from the [available ciphersuites](https://github.com/hashicorp/go-secure-stdlib/blob/tlsutil/v0.1.1/tlsutil/tlsutil.go#L26-L53)

When deploying with vault-helm, these can be set using the [injector.extraEnvironmentVariables](https://www.vaultproject.io/docs/platform/k8s/helm/configuration#extraenvironmentvars) chart option. 

Requires https://github.com/hashicorp/vault-k8s/pull/303 since this PR assumes [tls `Config.PreferServerCipherSuites` is ignored/deprecated](https://pkg.go.dev/crypto/tls#Config) as it is in go 1.17.

### DIFF
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-secure-stdlib/tlsutil"
 	agentInject "github.com/hashicorp/vault-k8s/agent-inject"
 	"github.com/hashicorp/vault-k8s/helper/cert"
 	"github.com/hashicorp/vault-k8s/leader"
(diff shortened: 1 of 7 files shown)


subcommand/injector/flags.go
```diff
@@ -13,8 +13,10 @@ import (
 )
 
 const (
-	DefaultLogLevel  = "info"
-	DefaultLogFormat = "standard"
+	DefaultLogLevel                    = "info"
+	DefaultLogFormat                   = "standard"
+	defaultTLSMinVersion               = "tls12"
```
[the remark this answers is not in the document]
kalafut: Just saw your comment elsewhere that at least back to K8s 1.14 there were no issues so perhaps it’s fine as is.

subcommand/injector/command.go
```diff
@@ -206,11 +211,28 @@ func (c *Command) Run(args []string) int {
 	}
 
 	var handler http.Handler = mux
+	minTLSVersion, ok := tlsutil.TLSLookup[c.flagTLSMinVersion]
+	if !ok {
+		c.UI.Error(fmt.Sprintf("Failed to parse minimum TLS version %q", c.flagTLSMinVersion))
```
benashz, line 216: The error message mentions parsing, but really we received an unsupported TLS version.

subcommand/injector/command.go
```diff
@@ -206,11 +211,28 @@ func (c *Command) Run(args []string) int {
...
+		Handler: handler,
+		TLSConfig: &tls.Config{
+			GetCertificate:           c.getCertificate,
+			MinVersion:               minTLSVersion,
+			PreferServerCipherSuites: c.flagTLSPreferServerCipherSuites,
+		},
+		ErrorLog: logger.StandardLogger(&hclog.StandardLoggerOptions{ForceLevel: hclog.Error}),
+	}
+	if len(ciphers) > 0 {
+		server.TLSConfig.CipherSuites = ciphers
```
benashz, line 235: nit: I feel like this change warrants factoring out the `TLSConfig{}` setup from `Server{}`

subcommand/injector/flags.go
```diff
@@ -13,8 +13,10 @@ import (
...
+	defaultTLSPreferServerCipherSuites = true
```
benashz, line 19: 👍

benashz (commented):
Only a few minor nits, otherwise 👍

subcommand/injector/flags.go
calvn: Is this flag/option needed? Reading from https://pkg.go.dev/crypto/tls#Config, it looks like this deprecated and is now completely ignored.

tvoran: Right, we can remove the option if we update our builds to go 1.17; they're on 1.16 currently.

tvoran: main is now building with go 1.17. I merged that into this branch, and removed the `tls_prefer_server_cipher_suites` option.

subcommand/injector/flags.go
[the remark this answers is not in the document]
tvoran: I did some simple tests on GKE and EKS with the default of TLS 1.2 and it all seemed to work.

subcommand/injector/flags.go
```diff
@@ -160,6 +167,11 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagResourceLimitMem, "memory-limit", agent.DefaultResourceLimitMem,
 		fmt.Sprintf("Memory resource limit set in injected containers. Defaults to %s", agent.DefaultResourceLimitMem))
 
+	c.flagSet.StringVar(&c.flagTLSMinVersion, "tls-min-version", defaultTLSMinVersion,
+		fmt.Sprintf(`Minimum supported version of TLS. Defaults to %s. Accepted values are "tls10", "tls11", "tls12" or "tls13"`, defaultTLSMinVersion))
+	c.flagSet.StringVar(&c.flagTLSCipherSuites, "tls-cipher-suites", "",
+		"Comma-separated list of supported cipher suites")
```
tomhjp, line 173: Is it worth mentioning that the order will not be respected? I think Golang will automatically reorder right? Maybe that's quite an advanced detail to include in quite high level help text though.

benashz: Perhaps, but it is really a matter of which go version was used during the build. I suppose that was one advantage of the previous flag. The server preferred order would be enabled regardless of version of go that was used.

subcommand/injector/flags.go
```diff
@@ -160,6 +167,11 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagResourceLimitMem, "memory-limit", agent.DefaultResourceLimitMem,
 		fmt.Sprintf("Memory resource limit set in injected containers. Defaults to %s", agent.DefaultResourceLimitMem))
 
+	c.flagSet.StringVar(&c.flagTLSMinVersion, "tls-min-version", defaultTLSMinVersion,
+		fmt.Sprintf(`Minimum supported version of TLS. Defaults to %s. Accepted values are "tls10", "tls11", "tls12" or "tls13"`, defaultTLSMinVersion))
```
benashz, line 171: I wonder if we should take these values from `tlsutil.TLSLookup` rather than hard coding them here? When tls14 is supported then that would be a valid as in input but not documented in the help.

subcommand/injector/command.go
```diff
@@ -265,6 +274,28 @@ func (c *Command) handleReady(rw http.ResponseWriter, req *http.Request) {
 	rw.WriteHeader(204)
 }
 
+func (c *Command) makeTLSConfig() (*tls.Config, error) {
```
benashz, line 277: 👍

benashz (approved):
LGTM!

subcommand/injector/flags.go
tvoran: Yeah, I like that idea, added in 892ba54. It comes out looking like this:

```
  -tls-min-version=<string>
     Minimum supported version of TLS. Defaults to tls12. Accepted
     values are tls10, tls11, tls12, tls13.
```

subcommand/injector/flags.go
```diff
@@ -160,6 +169,17 @@ func (c *Command) init() {
...
+	tlsVersions := []string{}
+	for v := range tlsutil.TLSLookup {
+		tlsVersions = append(tlsVersions, v)
+	}
+	sort.Strings(tlsVersions)
+	tlsStr := strings.Join(tlsVersions, ", ")
+	c.flagSet.StringVar(&c.flagTLSMinVersion, "tls-min-version", defaultTLSMinVersion,
+		fmt.Sprintf(`Minimum supported version of TLS. Defaults to %s. Accepted values are %s.`, defaultTLSMinVersion, tlsStr))
+	c.flagSet.StringVar(&c.flagTLSCipherSuites, "tls-cipher-suites", "",
+		"Comma-separated list of supported cipher suites for TLS 1.0-1.2")
```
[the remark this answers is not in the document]
benashz: Very interesting...

In that case, perhaps rephrasing it a bit to be:

```suggestion
		"Comma-separated list of supported cipher suites (ignored for tlsv3 and higher)")
```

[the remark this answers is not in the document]
tvoran: Thanks, though I think we'll keep this phrasing for now.

PR #309, opened 2021-12-01
Adding github workflow for CI tests

tvoran:
Adds a github workflow for running the CI tests (build and unit tests).

Leaves the circleci config in place, until we set it to stop building there.


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.2
```
tomhjp, line 26: Might as well go with 1.17.4?

tomhjp (approved):

.github/workflows/tests.yaml
tvoran: Yep, done in 6fd14e3.

PR #316, opened 2022-01-20
Changelog update for v0.14.2

tvoran:
Updating the changelog with the changes in v0.14.2


kalafut (approved):
Interesting about the date…

tvoran:
> Interesting about the date…

Yeah, I suspect it was using UTC.

PR #324, opened 2022-03-04
Use CRT for releases

tomhjp:
* New `build` job to make artifacts
* CRT config in `.release/`
* Define Go version in one place for the whole repo at `.go-version`
* Add manual dispatch for `build` job that lets us define the release version at release time
* Combine dev and prod Docker files into one multi-stage Docker file

UBI images are currently out of scope, but to be addressed in a later PR.


Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Create a non-root user to run the software.
+RUN addgroup vault && \
+    adduser -S -G vault vault
+
+# Set up certificates, base tools, and software.
+RUN set -eux && \
+    apk add --no-cache ca-certificates libcap su-exec iputils
+
+COPY dist/$TARGETOS/$TARGETARCH/vault-k8s /bin/
```
tvoran, line 43: What's putting the binary in this directory structure? I see the build step zip's the binary into a file:

https://github.com/hashicorp/vault-k8s/pull/324/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R87-R90

But that doesn't match this directory structure.

tvoran (approved):
This seems to be working, though I'm not sure I understand how all the pieces fit together yet. Looks like we can remove the `build/docker/{Dev,Release}.dockerfile`, and the `prod-image` and `xc-prod-image` make targets too?

Dockerfile
tomhjp: tbh I couldn't have given you a full answer until I looked it up just now, but the short answer is that the `hashicorp/actions-docker-build` action we use extracts the built GitHub artifact into this location.

A digest_inputs script gets run first: https://github.com/hashicorp/actions-docker-build/blob/5f52c4cae49d2f379cd1fcb4a4f09b55f97a3f06/action.yml#L109

That sets a ZIP_LOCATION env var which matches our COPY directory here: https://github.com/hashicorp/actions-docker-build/blob/main/scripts/digest_inputs#L78

We then download the zip artifact into that folder: https://github.com/hashicorp/actions-docker-build/blob/main/action.yml#L119

And finally we just extract the binary into the same directory as the zip: https://github.com/hashicorp/actions-docker-build/blob/main/scripts/extract_product_zip#L13

For reference, that -j option on unzip:

> junk paths. The archive's directory structure is not recreated; all files are deposited in the extraction directory (by default, the current one).

PR #326, opened 2022-03-10
Build with go 1.17.8

ericofusco:
Build with go 1.17.8 

Release history: https://go.dev/doc/devel/release#go1.17.minor

Not sure if this is the only change required, let me know if more changes are required.


tvoran (approved):
Thanks! I think you got everything.

PR #327, opened 2022-03-21
Release 0.15.0

tvoran:
Update images and deps and default vault version for 0.15.0 release


jasonodonnell (approved):

Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN addgroup vault && \
 
 # Set up certificates, base tools, and software.
 RUN set -eux && \
+    apk update && apk upgrade libretls && \
```
kalafut, line 41: Should we add a ticket to remind us to check whether we can remove this (i.e. base image is updated)?

PR #329, opened 2022-03-30
Add agent metrics annotations & env vars

BrandonS05:
Now that Vault Agent metrics have been released in v1.10.0 (https://github.com/hashicorp/vault/pull/13675), the injector needs a way to enable and configure them. There is a strong need for those using the Vault Injector in production to have insights on Vault authentication/ connection issues from the injected agent. This pull request addresses #331.

A few notes on this pull request:
- Annotations have not been added for setting up server-side TLS on the exposed metrics address; this would be a great feature to add later. If there is a strict requirement for TLS right now, this can still be achieved either via a service mesh or by setting the metrics address to localhost for a custom application to consume and expose on its own terms.
- Until https://github.com/hashicorp/vault/issues/8953 is resolved, an empty cache block will be added as a temporary workaround to the configuration if a cache configuration isn't provided already
- The "vault agent -config" command currently adds cache endpoints to all listeners regardless of their purpose, which can create an unintentionally insecure configuration if a user only wants to expose a metrics endpoint externally. Until this issue is resolved on the Vault Agent, a temporary safeguard has been implemented that prevents users with auto-auth configured on the cache from also configuring an external metrics listening address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Unreleased
 
+## 0.16.0 (March 30, 2022)
+
+Features:
+* New environment variable and annotation to configure metrics listener address for Vault agent
+* New environment variable and annotation to enable prometheus metrics and configure retention time for Vault agent
+
 ## 0.15.0 (March 21, 2022)
 
 Features:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,7 @@
 ## 0.16.0 (March 30, 2022)
 
 Features:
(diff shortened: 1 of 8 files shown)


tvoran:
Hi there, welcome to the HashiCorp Vault community! Please note that it's best if PRs are associated with an issue, either an existing one or a new issue, so that the community and the engineers can discuss implementation details, strategy, and where the request fits into an existing roadmap. Additionally, we can't do much more with this PR until the CLA is signed. Please feel free to create or link an existing issue to this PR. Thanks!

PR #330, opened 2022-04-01
add agent-enable-quit annotation

tvoran:
New annotation to enable the [/agent/v1/quit endpoint](https://www.vaultproject.io/docs/agent#quit) on an injected agent. This option defaults to `false`, and will be set if `true` on the existing cache listener, or a new localhost listener with a basic cache stanza configured since that's required for an agent listener.

The [agent-cache-listener-port](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-cache-listener-port) annotation can be used to change the port.

Example Job:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: quit

spec:
  template:
    metadata:
      annotations:
        vault.hashicorp.com/role: "internal-app"
        vault.hashicorp.com/agent-inject-secret-database-config.txt: "internal/data/database/config"
        vault.hashicorp.com/agent-inject-template-database-config.txt: |
          {{- with secret "internal/data/database/config" -}}
          postgresql://{{ .Data.data.username }}:{{ .Data.data.password }}@postgres:5432/wizard
          {{- end -}}
        vault.hashicorp.com/agent-enable-quit: "true"
    spec:
      serviceAccountName: internal-app
      containers:
      - name: nginx
        image: nginx
        command: ["/bin/bash", "-c"]
        args:
          - |
            sleep 10
            curl -X POST http://localhost:8200/agent/v1/quit

      restartPolicy: Never
  backoffLimit: 4
```

### DIFF
--- a/agent-inject/agent/config.go
+++ b/agent-inject/agent/config.go
@@ -215,24 +215,21 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		}
 	}
 
-	// If EnableQuit has been set, set it on the listener. If a listener hasn't
-	// been defined, set it on a new one. Also add a simple cache stanza since
-	// that's required for an agent listener.
-	if a.EnableQuit != nil {
+	// If EnableQuit is true, set it on the listener. If a listener hasn't been
+	// defined, set it on a new one. Also add a simple cache stanza since that's
+	// required for an agent listener.
+	if a.EnableQuit {
 		if len(config.Listener) > 0 {
 			config.Listener[0].AgentAPI = &AgentAPI{
-				EnableQuit: *a.EnableQuit,
+				EnableQuit: a.EnableQuit,
 			}
 		} else {
-			// Only setup a new listener if EnableQuit is true
-			if *a.EnableQuit {
-				config.Listener = makeCacheListener(a.VaultAgentCache.ListenerPort)
-				config.Listener[0].AgentAPI = &AgentAPI{
-					EnableQuit: *a.EnableQuit,
-				}
+			config.Listener = makeCacheListener(a.VaultAgentCache.ListenerPort)
+			config.Listener[0].AgentAPI = &AgentAPI{
+				EnableQuit: a.EnableQuit,
 			}
 		}
-		if *a.EnableQuit && config.Cache == nil {
+		if config.Cache == nil {
 			// Cache is required for an agent listener
 			config.Cache = &Cache{}
 		}
--- a/agent-inject/agent/config.go
+++ b/agent-inject/agent/config.go
@@ -95,7 +95,7 @@ type AgentAPI struct {
 
 // Cache defines the configuration for the Vault Agent Cache
 type Cache struct {
-	UseAutoAuthToken string        `json:"use_auto_auth_token"`
+	UseAutoAuthToken string        `json:"use_auto_auth_token,omitempty"`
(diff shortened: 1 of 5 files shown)


agent-inject/agent/config.go
```diff
@@ -215,9 +215,42 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		}
 	}
 
+	// If EnableQuit has been set, set it on the listener. If a listener hasn't
+	// been defined, set it on a new one. Also add a simple cache stanza since
+	// that's required for an agent listener.
+	if a.EnableQuit != nil {
+		if len(config.Listener) > 0 {
+			config.Listener[0].AgentAPI = &AgentAPI{
```
tomhjp, line 223: In the vast majority of cases, there will only be 1 listener, but should we maybe set it on all listeners? Just choosing the first one seems surprising, and for users who want to finesse which listeners it is and isn't on, I think it's ok to not give them a dedicated annotation for that but let them fully define their own config.

agent-inject/agent/config.go
```diff
@@ -215,9 +215,42 @@ func (a *Agent) newConfig(init bool) ([]byte, error) {
 		}
 	}
 
+	// If EnableQuit has been set, set it on the listener. If a listener hasn't
+	// been defined, set it on a new one. Also add a simple cache stanza since
+	// that's required for an agent listener.
+	if a.EnableQuit != nil {
```
tomhjp, line 221: Do we need to do anything if it's false, given it defaults to false?

tvoran: Good question, and I went back and forth on this quite a bit. It would certainly be simpler to only set the enable_quit config if the annotation has been set to true, and maybe that's the best to do for now.

Currently `EnableQuit` is setup as a pointer so the config is set only if the annotation was specified (true or false), but since it defaults to false in the agent, just omitting the config setting here should be sufficient. And it's probably very unlikely that we'd ever default to enabling the quit endpoint on the vault agent side.

agent-inject/agent/config.go
tvoran: I don't think the injector can currently create a config with more than one listener, can it? I think the only way to inject an agent that has multiple listener is to specify the full config in a configmap, and in that case [this code wouldn't be triggered](https://github.com/hashicorp/vault-k8s/blob/afa2571312f017efb944a504abc9603a1f1343b3/agent-inject/agent/container_env.go#L49-L61).

But I can add a loop since it does look a little jarring to only reference listener[0] here.

agent-inject/agent/config.go
tomhjp: Yeah I think that seems like a safe bet for now 👍

agent-inject/agent/config.go
tomhjp: Oh I forgot about that - makes sense.

tomhjp (approved):

agent-inject/agent/config.go
tvoran: Updated in [3b474e8](https://github.com/hashicorp/vault-k8s/pull/330/commits/3b474e85c1d97acec915e5928f435992b06515f1) to only set the config if EnableQuit is true.

tomhjp (approved):
LGTM 👍

PR #333, opened 2022-04-08
add go-max-procs annotation

packageman:
When Golang program started in a container, the P (from Golang GMP model) will be set to the number of CPU cores of host instead of container's CPU limit (cgroups).

Assuming vault-agent sidecar is deployed on a 16core64g machine, the P of vault-agent will be set to the number of CPU cores by default, which is 16, and the available CPU of the vault-agent is 500m ([helm chart default setting](https://github.com/hashicorp/vault-helm/blob/main/values.yaml#L67)), which will cause the CPU to be used too high constantly. Solved by explicitly setting `GOMAXPROCS` to 1.

Therefore, add annotation `vault.hashicorp.com/go-max-procs` to support configuring P's number accordingly.

- P = 16

perf-vault-agent

image

- P = 1

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Features:
 * Add agent-enable-quit annotation [GH-330](https://github.com/hashicorp/vault-k8s/pull/330)
+* Add go-max-procs annotation [GH-333](https://github.com/hashicorp/vault-k8s/pull/333)
 
 ## 0.15.0 (March 21, 2022)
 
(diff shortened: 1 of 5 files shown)


tvoran:
Thanks for the analysis here! This looks pretty straightforward, but would you mind testing for that variable in `agent-inject/agent/container_env_test.go`?

packageman:
@tvoran tests added.

swenson (approved):
LGTM. I pulled this down and tested it, and it seems to work as expected. Thanks!

PR #336, opened 2022-04-22
VAULT-2230 Only update webhooks CA bundle when needed

### LINKED ISSUE #332: Admission controller registration is invoked every second
**Describe the bug**
The `certWatcher` is calling kubernetes api once per second to patch the mutating webhook config, even when no certificate change takes place. This results in a lot of log spam and extra cloud provider charges (even if they may be relatively small).

https://github.com/hashicorp/vault-k8s/blob/c25fcef11201a25d74f73cf62f2006b05997f826/subcommand/injector/command.go#L328

The certWatcher should notice nothing has changed since its last cert update and avoid invoking the api unnecessarily.

**To Reproduce**
1. Run the sidecar
2. Examine activity such as kubernetes controller logs

swenson:
Instead of every second. Specifically, we now only update the webhooks
bundle if:

* we have generated a certificate and the CA bundle does not match the
  current webhook CA bundle; this will also happen if a new leader is
  elected, or
* the webhook itself was changed (e.g., by a helm upgrade), checked by
  `Watch`ing the webhooks.

In any case, will only send the `Patch` if the CA bundle on the
webhook no longer matches our own CA bundle.

This was tested by performing a `helm upgrade` with a webhook change and
verifying that the webhook change was noticed and the CA bundle was
patched, and also by setting the certificate lifetimes very low and
making sure that the CA bundle was *not* patched (since the CA did not
change when the certificate was regenerated).

Fixes #332.

### DIFF
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -326,6 +326,8 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 		panic(err)
 	}
 
+	defaultLoopTime := 1 * time.Hour // update after this amount of time even if nothing has happened
+	timer := time.NewTimer(defaultLoopTime)
 	expBackoff := backoff.NewExponentialBackOff()
 
 	for {
@@ -341,12 +343,23 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 		case <-ctx.Done():
 			// Quit
 			return
+
+		case <-timer.C:
+			// we are told to retry or periodically update
 		}
+
+		// clear the timer
+		if !timer.Stop() {
+			<-timer.C
+		}
+
 		err := c.updateWebhook(ctx, clientset, bundle, webhooksCache, leaderElector, log)
 		if err != nil {
-			time.Sleep(expBackoff.NextBackOff())
+			// retry after a delay
+			timer.Reset(expBackoff.NextBackOff())
 		} else {
 			expBackoff.Reset()
+			timer.Reset(defaultLoopTime)
 		}
 	}
 }
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -368,7 +368,7 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 					err))
 				continue
 			}
-			if len(currentBundles) > 0 && bytes.Equal(currentBundles[0], bundle.CACert) {
+			if len(currentBundles) == 0 || !bytes.Equal(currentBundles[0], bundle.CACert) {
 				err = c.updateCABundle(ctx, &bundle, clientset, adminAPIVersion)
 				if err != nil {
 					c.UI.Error(fmt.Sprintf(
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -16,6 +16,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-secure-stdlib/tlsutil"
 	agentInject "github.com/hashicorp/vault-k8s/agent-inject"
@@ -174,7 +175,7 @@ func (c *Command) Run(args []string) int {
 	certCh := make(chan cert.Bundle)
 	certNotify := cert.NewNotify(ctx, certCh, certSource, logger.Named("notify"))
 	go certNotify.Run()
-	go c.certWatcher(ctx, certCh, clientset, leaderElector, namespace, logger.Named("certwatcher"))
+	go c.certWatcher(ctx, certCh, clientset, leaderElector, logger.Named("certwatcher"))
 
 	// Build the HTTP handler and server
 	injector := agentInject.Handler{
@@ -317,19 +318,21 @@ func getAdminAPIVersion(ctx context.Context, clientset *kubernetes.Clientset) (s
 	return adminAPIVersion, err
 }
 
-func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, clientset *kubernetes.Clientset, leaderElector leader.Elector, namespace string, log hclog.Logger) {
+func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, clientset *kubernetes.Clientset, leaderElector leader.Elector, log hclog.Logger) {
 	var bundle cert.Bundle
-	webhooksCache, webhooksWatcher, err := watchWebhooks(ctx, clientset, namespace)
+	webhooksCache, webhooksWatcher, err := watchWebhooks(ctx, clientset)
 	if err != nil {
 		log.Error("Error watching webhooks", "error", err)
 		panic(err)
 	}
 	tryImmediately := false
 
+	expBackoff := backoff.NewExponentialBackOff()
+
 	for {
 		if tryImmediately {
 			// avoid tight loop on failure
-			time.Sleep(1 * time.Second)
+			time.Sleep(expBackoff.NextBackOff())
 		} else {
 			select {
 			case bundle = <-ch:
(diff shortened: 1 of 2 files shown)


subcommand/injector/command.go
```diff
@@ -353,47 +358,139 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
+			// Check the current bundles and only update if necessary.
+			currentBundles, err := c.getCurrentCABundles(ctx, clientset, adminAPIVersion)
 			if err != nil {
 				c.UI.Error(fmt.Sprintf(
-					"Error updating MutatingWebhookConfiguration: %s",
+					"Error getting MutatingWebhookConfiguration: %s",
 					err))
 				continue
 			}
+			if len(currentBundles) > 0 && bytes.Equal(currentBundles[0], bundle.CACert) {
```
jasonodonnell, line 371: Should this be `!bytes.Equal(currentBundles[0], bundle.CACert)`?

swenson: Oops. Good catch!

subcommand/injector/command.go
```diff
@@ -353,47 +358,139 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
 }
 
+func watchWebhooks(ctx context.Context, clientset *kubernetes.Clientset, adminAPIVersion string) (watch.Interface, error) {
+	switch adminAPIVersion {
+	// AdmissionregistrationV1 is valid in k8s 1.16+
+	case "v1":
+		return clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Watch(ctx, metav1.ListOptions{})
+	// AdmissionregistrationV1beta1 is present in k8s 1.9 - 1.19
+	case "v1beta1":
+		return clientset.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Watch(ctx, metav1.ListOptions{})
```
tvoran, line 394: FWIW, I think we could drop support for k8s older than 1.16 in this, since that's the minimum version we now support in vault-helm: https://github.com/hashicorp/vault-helm/blob/main/Chart.yaml#L5

jasonodonnell (approved):
LGTM!

subcommand/injector/command.go
```diff
@@ -318,18 +320,21 @@ func getAdminAPIVersion(ctx context.Context, clientset *kubernetes.Clientset) (s
 
 func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, clientset *kubernetes.Clientset, leaderElector leader.Elector, adminAPIVersion string, log hclog.Logger) {
 	var bundle cert.Bundle
+	webhooksWatcher, err := watchWebhooks(ctx, clientset, adminAPIVersion)
+	if err != nil {
+		log.Warn("Error watching webhooks", "error", err)
+		webhooksWatcher = newTickerWatcher(1 * time.Second)
+	}
```
tvoran, line 327: This is falling back to the old method of patching once a second? (Just making sure I follow.)

subcommand/injector/command.go
```diff
@@ -353,47 +358,139 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
 		// Update the certificate
 		c.cert.Store(&crt)
 	}
 }
 
+func watchWebhooks(ctx context.Context, clientset *kubernetes.Clientset, adminAPIVersion string) (watch.Interface, error) {
+	switch adminAPIVersion {
+	// AdmissionregistrationV1 is valid in k8s 1.16+
+	case "v1":
+		return clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Watch(ctx, metav1.ListOptions{})
```
tvoran, line 391: This will return watch events for all mutating webhook configs on the system? Are there any ListOptions that could filter that list so it isn't triggered on non-vault-k8s webhook changes?

I'm also wondering if you considered setting up an informer here. IIRC a straight `Watch` isn't as robust when it comes to network interruptions (these discussions are a little old, so maybe my knowledge is outdated): 
* https://github.com/kubernetes-client/javascript/issues/377
* https://stackoverflow.com/questions/59544139/kubernetes-client-go-watch-interface-vs-cache-newinformer-vs-cache-newsharedi

Another option would be to poll against the informer cache's lister interface, as we do with the cert secret: https://github.com/hashicorp/vault-k8s/blob/0c69acbf33605f66cd1316fd9cd5380e7560f84d/helper/cert/source_gen.go#L263

subcommand/injector/command.go
swenson: Yep! I figured if something was wrong, like we don't have permission to WATCH that endpoint or something, go back to the old behavior.

subcommand/injector/command.go
swenson: Ah, perfect. I didn't know about informers. That sounds much more robust.

subcommand/injector/command.go
swenson: Cool. I'll go ahead and drop the "v1beta1" support.

swenson:
Thanks for the suggestions. This is ready for review again. I've switched to using an informer.

tvoran (approved):
Worked great for me when testing locally. Left a couple thoughts, but I think this is good.

subcommand/injector/command.go
```diff
@@ -316,29 +317,53 @@ func getAdminAPIVersion(ctx context.Context, clientset *kubernetes.Clientset) (s
...
-			// Bundle is updated, set it up
-
-		case <-time.After(1 * time.Second):
-			// This forces the mutating webhook config to remain updated
-			// fairly quickly. This is a jank way to do this and we should
-			// look to improve it in the future. Since we use Patch requests
-			// it is pretty cheap to do, though.
+		if tryImmediately {
+			// avoid tight loop on failure
+			time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
calvn: We could also wrap all of the logic that's after the `select` within this for loop to its own func which sends the error to an `errCh` that's also hooked into the select here (and we can sleep for 1s in there instead). It has the benefit of getting rid of the additional nesting and various `continue` paths that exists here.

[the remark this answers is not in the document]
swenson: :+1:

subcommand/injector/command.go
```diff
@@ -394,6 +399,50 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 	}
 }
 
+func watchWebhooks(ctx context.Context, clientset *kubernetes.Clientset, namespace string) (cache.Store, <-chan interface{}, error) {
+	factory := informers.NewSharedInformerFactoryWithOptions(clientset, 0, informers.WithNamespace(namespace))
+	webhooks := factory.Admissionregistration().V1().MutatingWebhookConfigurations()
```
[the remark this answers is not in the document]
swenson: You're right. It doesn't seem to matter. I'll remove it. :)

subcommand/injector/command.go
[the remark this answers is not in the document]
swenson: Re: `errCh`, it didn't make a lot of sense to add a separate channel for that, since we don't want to execute the update function as a goroutine (we could get into some race conditions).

But, it does make sense to move it to a separate function and simplify the backoff logic a little, so I did that. Thanks for the suggestion.

subcommand/injector/command.go
```diff
@@ -316,82 +318,131 @@ func getAdminAPIVersion(ctx context.Context, clientset *kubernetes.Clientset) (s
...
 			// Quit
 			return
 		}
-
-		crt, err := tls.X509KeyPair(bundle.Cert, bundle.Key)
+		err := c.updateWebhook(ctx, clientset, bundle, webhooksCache, leaderElector, log)
 		if err != nil {
-			log.Warn(fmt.Sprintf("Could not load TLS keypair: %s. Trying again...", err))
-			continue
+			time.Sleep(expBackoff.NextBackOff())
```
calvn, line 347: This will sleep for the backoff duration once, but it will block on the next for loop run until the next case is selected, which may be munch longer than the backoff duration here.

swenson: Great point. We do want to retry here. I'll rework the loop.

swenson:
@calvn let me know if you want to take another look

calvn (approved):

swenson:
Thanks!

PR #341, opened 2022-05-06
Support min/max backoff annotations and env vars

swenson:
When auto auth fails, the agent will automatically retry with an
exponential backoff starting from 1 second up to 5 minutes. Those values
can be overridden in Vault itself. Here, we wire up agent to pass min
and max backoff values from its environment variables or the pod's
annotations.

* If the `AGENT_INJECT_AUTH_MIN_BACKOFF` or `AGENT_INJECT_MAX_BACKOFF`
  environment variables are set on the agent, these will be used as the
  default values.
* If a pod has the annotations `vault.hashicorp.com/auth-min-backoff` or
  `vault.hashicorp.com/auth-max-backoff` set, these will be used as the
  min and max backoff values. These will override any defaults from the
  agent environment variables.

The environment variables can be set via helm as well via, for example,

```sh
helm install vault hashicorp/vault \
  --set "injector.extraEnvironmentVars.AGENT_INJECT_AUTH_MIN_BACKOFF=4s" \
  --set "injector.extraEnvironmentVars.AGENT_INJECT_AUTH_MAX_BACKOFF=5s"
```

This was tested via `helm` and by setting the pod annotations and
confirming that the Vault agent picked up the settings when auto auth
is failing.

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -5,6 +5,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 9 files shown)


agent-inject/agent/annotations.go
```diff
@@ -451,6 +462,30 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval] = cfg.StaticSecretRenderInterval
 	}
 
+	if minBackoffString, ok := pod.ObjectMeta.Annotations[AnnotationAgentAuthMinBackoff]; ok {
+		if minBackoffString != "" {
+			_, err := time.ParseDuration(minBackoffString)
+			if err != nil {
+				return fmt.Errorf("error parsing min backoff as duration: %v", err)
+			}
```
tvoran, line 470: I wonder if we need to parse the backoff formats here at all, since Vault Agent will do its own checking? Similar to how we just pass through StaticSecretRenderInterval above.

tvoran (approved):
Looks and works great when I tested it! Just a question about how much parsing we need to do of the backoff values vs letting vault agent do its own checking.

agent-inject/agent/annotations.go
swenson: I don't know that we need to, but I thought it might be nice to have a quick check to prevent some other weird error if something got through.

swenson:
Thanks!

Issue #343, opened 2022-05-16
"Could not find webhook config in cache"-loop leading to TLS handshake: no certificate available error


lukaalba:
**Describe the bug**
We use the vault-agent-injector with a certificate generated by cert-manager. With the update to 0.16.0, all replicas of the vault-agent-injector produce the following error:

```Registering telemetry path on "/metrics"
Listening on ":8080"...
2022-05-16T10:23:48.956Z [INFO]  handler: Starting handler..
2022-05-16T10:23:49.056Z [INFO]  handler.certwatcher: Webhooks changed. Updating certs...
2022-05-16T10:23:49.056Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:23:49.056Z [INFO]  handler.certwatcher: Webhooks changed. Updating certs...
2022-05-16T10:23:49.056Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:23:49.056Z [INFO]  handler.certwatcher: Webhooks changed. Updating certs...
2022-05-16T10:23:49.056Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:23:49.056Z [INFO]  handler.certwatcher: Webhooks changed. Updating certs...
2022-05-16T10:23:49.056Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:23:49.056Z [INFO]  handler.certwatcher: Webhooks changed. Updating certs...
2022-05-16T10:23:49.056Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:23:49.056Z [INFO]  handler.certwatcher: Webhooks changed. Updating certs...
2022-05-16T10:23:49.056Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:23:49.206Z [INFO]  handler.certwatcher: Updated certificate bundle received. Updating certs...
2022-05-16T10:23:49.206Z [WARN]  handler.certwatcher: Could not find webhook config in cache. Trying again...: item=<nil>
2022-05-16T10:24:00.337Z [ERROR] handler: http: TLS handshake error from 172.16.1.1:34462: no certificate available
[...]

PR #344, opened 2022-05-16
Don't update the mutating webhook if we aren't managing it automatically

swenson:
If the webhook isn't being managed by us, then `c.flagAutoName` will be empty, so we shouldn't try to update the webhook ourselves.

### DIFF
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -368,20 +368,34 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
 	}
 }
 
-func (c *Command) updateWebhook(ctx context.Context, clientset *kubernetes.Clientset, bundle cert.Bundle, webhooksCache cache.Store, leaderElector leader.Elector, log hclog.Logger) error {
+func (c *Command) fetchWebhook(ctx context.Context, clientset *kubernetes.Clientset) (*adminv1.MutatingWebhookConfiguration, error) {
+	return clientset.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(ctx, c.flagAutoName, metav1.GetOptions{})
+}
+
+func (c *Command) getWebhookCached(webhooksCache cache.Store) (*adminv1.MutatingWebhookConfiguration, error) {
 	item, exists, err := webhooksCache.GetByKey(c.flagAutoName)
 	if !exists {
-		log.Warn("Could not find webhook config in cache. Trying again...", "item", item)
-		return fmt.Errorf("could not find webhook config in cache")
+		return nil, fmt.Errorf("webhook config not cached")
 	} else if err != nil {
-		log.Warn(fmt.Sprintf("Could not find webhook config in cache: %s. Trying again...", err))
-		return err
+		return nil, err
 	}
 	config, ok := item.(*adminv1.MutatingWebhookConfiguration)
 	if !ok {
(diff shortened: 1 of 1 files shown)


swenson:
I tested this with cert-manager, and confirmed that:

* 0.14.2 works
* 0.16.0 fails with an authentication problem, as per https://github.com/hashicorp/vault-k8s/issues/343
* The code in this branch works as expected

subcommand/injector/command.go
```diff
@@ -368,20 +368,25 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
 	} else if err != nil {
-		log.Warn(fmt.Sprintf("Could not find webhook config in cache: %s. Trying again...", err))
-		return err
+		return nil, err
 	}
 	config, ok := item.(*adminv1.MutatingWebhookConfiguration)
 	if !ok {
-		log.Error("Got unknown object from cache; expected &MutatingWebhookConfiguration", "item", item)
+		return nil, fmt.Errorf("got unknown object from cache; expected &MutatingWebhookConfiguration but got %T", item)
 	}
```
tvoran, line 385: As I was poking around in this, I noticed that if we were to use the Lister() interface from the MutatingWebhookConfigurationInformer instead of Informer().GetStore(), that `Lister().Get` would give back the typed adminv1.MutatingWebhookConfiguration object without needing to assert it here. (Just a thought.)

subcommand/injector/command.go
```diff
@@ -368,20 +368,25 @@ func (c *Command) certWatcher(ctx context.Context, ch <-chan cert.Bundle, client
...
+	return config, nil
+}
 
+func (c *Command) updateWebhook(ctx context.Context, clientset *kubernetes.Clientset, bundle cert.Bundle, webhooksCache cache.Store, leaderElector leader.Elector, log hclog.Logger) error {
```
tvoran, line 389: Could we rename this function to something like updateCerts, or updateWebhookAndCertStore? Just to denote that it's doing more than just updating the webhook config.

subcommand/injector/command.go
swenson: Oh cool. I'll switch to that.

subcommand/injector/command.go
swenson: :+1:

subcommand/injector/command.go
swenson: Oh, but then we wouldn't be reading from the cache, which we know is up-to-date, but then fetching it again in the `Lister()`, which isn't ideal. (It could be out of sync even.) I think it's probably best to stick with the cache.

tvoran: Oh? I thought the Lister was giving back essentially an indexed view on the cache.

https://github.com/kubernetes/client-go/blob/60a0346672170c8e9f65795fb33e41527980c583/informers/admissionregistration/v1/mutatingwebhookconfiguration.go#L87

vs

https://github.com/kubernetes/client-go/blob/60a0346672170c8e9f65795fb33e41527980c583/informers/admissionregistration/v1/mutatingwebhookconfiguration.go#L87

tvoran (approved):

subcommand/injector/command.go
swenson: Ah, you appear correct. Will fix.

swenson:
Thanks!

PR #345, opened 2022-05-17
Added support to configure default vault namespace on the agent config

flaviolemos78:
At the moment the only option to configure vault enterprise namespace is through the annotation `vault.hashicorp.com/namespace` at Pod level. 
This presents some autonomy challenges for cluster/vault administrators. Example: 

- When migrating to Vault enterprise namespaces, administrators aren't able to seamless change the injector vault until the last "workload" add the namespace annotation....

This PR adds the ability to define a default vault namespace on the agent config. If the vault namespace isn't defined at Pod, then the configured vault namespace will be used, otherwise the Pod annotation value is used


tvoran (approved):
Thanks! This is a nice addition, and tested great for me.

flaviolemos78:
Hi @tvoran, do you know when will be the next release with this?

bradfordwagner:
@flaviolemos78  - I'll follow this up with a MR against the helm chart.

PR #346, opened 2022-05-17
Fixed certificate assignment when not using auto mode

flaviolemos78:
When the agent is not running on "auto mode" the agent is failing to proper assign the TLS certificate, causing TLS handshake errors. This is happening because before setting the certificate, the code is looking for the webhook on a `webhookcache`  and since the webhook name is identified by the flag `flagAutoName` and the flag is empty, the code is constantly returning before actually setting the certificate.


swenson:
Thanks for this! We found it #344 as well, which we've gone ahead and merged.

Issue #258, opened 2021-05-21
Injector mutating webhook possibly fatal for a cluster


sttts:
https://github.com/hashicorp/vault-k8s/blob/master/deploy/injector-mutating-webhook.yaml#L22 misses label selectors to exclude the webhook itself. This can be fatal for a cluster as it will stop creating pods.

joeyslalom:
Was bitten by this one.  I set `injector.failurePolicy=Fail`, then tried to apply the helm chart but got deployment errors:

`Error creating: Internal error occurred: failed calling webhook "vault.hashicorp.com": Post "https://vault-agent-injector-svc.vault.svc:443/mutate?timeout=4s": no endpoints available for service "vault-agent-injector-svc"   FailedCreate`

I resolved this by using `injector objectSelector` as follows:
```
  objectSelector:
    matchExpressions:
    - key: app.kubernetes.io/name
      operator: NotIn
      values: ["vault-agent-injector"]
```

nyetwurk:
What is the status of this ticket?

I see https://github.com/hashicorp/vault-k8s/blob/286c36fcff05636581cdedcc4b2686da138ca4bc/deploy/injector-mutating-webhook.yaml#L26 has not changed.

nyetwurk:
What is the status of this issue? I see https://github.com/hashicorp/vault-k8s/blob/286c36fcff05636581cdedcc4b2686da138ca4bc/deploy/injector-mutating-webhook.yaml#L26 has not changed.

nyetwurk:
```yaml
    objectSelector:
      matchExpressions:
      - key: app.kubernetes.io/name
        operator: NotIn
        values: ["vault-injector"]
```
LGTM

PR #351, opened 2022-05-23
Update example injector mutating webhook config to exclude agent pod

swenson:
- Original issue: https://github.com/hashicorp/vault-k8s/issues/258
- Helm fix: https://github.com/hashicorp/vault-helm/pull/736

### DIFF
--- a/deploy/injector-mutating-webhook.yaml
+++ b/deploy/injector-mutating-webhook.yaml
@@ -24,5 +24,9 @@ webhooks:
         apiVersions: ["v1"]
         resources: ["deployments", "jobs", "pods", "statefulsets"]
     namespaceSelector: {}
-    objectSelector: {}
+    objectSelector:
+      matchExpressions:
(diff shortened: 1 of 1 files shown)


deploy/injector-mutating-webhook.yaml
```diff
@@ -24,5 +24,9 @@ webhooks:
         apiVersions: ["v1"]
         resources: ["deployments", "jobs", "pods", "statefulsets"]
     namespaceSelector: {}
-    objectSelector: {}
+    objectSelector:
+      matchExpressions:
+      - key: app.kubernetes.io/name
+        operator: NotIn
+        values: ["vault-agent-injector"]
```
tvoran, line 31: It looks like pods deployed here use "vault-injector" instead of "vault-agent-injector".

```suggestion
        values: ["vault-injector"]
```

tvoran (approved):

tomhjp (approved):

PR #353, opened 2022-05-23
ConfigMap with missing vault section should default to env vars

### LINKED ISSUE #300: Injector don't set external vault address when using ConfigMap
**Describe the bug**
When injector has been configured with external Vault address and user mounts Vault HCL ConfigMap without `vault` configuration section the authentication runs agains localhost instead of external address.

**To Reproduce**
1. Deploy the vault injector via helm chart with `externalVaultAddr` value configured
2. Create ConfigMap without optional `vault` section that specifies external address
3. Deploy the application with `vault.hashicorp.com/agent-configmap` annotation

**Expected behavior**
Mutation webhook injects VAULT_ADDR environment variable into the Pod so both HCL and annotations work with pre-configured external server. Current functionality hardcodes vault address into VAULT_CONFIG / config.json only when using annotations to retrieve the secrets.

**Environment**
* Kubernetes version: 
  * Distribution or cloud vendor (OpenShift, EKS, GKE, AKS, etc.): AKS, GKE, minikube
  * Other configuration options or runtime services (istio, etc.): Spark Operator, Istio
* vault-k8s version: hashicorp/vault:1.8.2

swenson:
When using a ConfigMap to hold the agent config, it is natural to omit
the `vault` section, since the config may be hard-coded and won't know
necessarily know the correct address for the Vault server.

In that case, we should set any known information about the Vault server
(address, cert, etc.) in environment variables passed to the agent so
that it can connect to the correct Vault server.

Fixes #300

### DIFF
--- a/agent-inject/agent/container_env.go
+++ b/agent-inject/agent/container_env.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"encoding/base64"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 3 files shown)


agent-inject/agent/container_env.go
```diff
@@ -64,6 +65,48 @@ func (a *Agent) ContainerEnvVars(init bool) ([]corev1.EnvVar, error) {
...
+		envs = append(envs, corev1.EnvVar{
+			Name:  "VAULT_SKIP_VERIFY",
+			Value: strconv.FormatBool(a.Vault.TLSSkipVerify),
+		})
+		if a.Vault.TLSServerName != "" {
+			envs = append(envs, corev1.EnvVar{
+				Name:  "VAULT_TLS_SERVER_NAME",
+				Value: a.Vault.TLSServerName,
+			})
+		}
```
calvn, line 109: Should we also include `VAULT_NAMESPACE`?

tomhjp: Good question. Namespace isn't part of the `vault` config block, only `auto_auth`. And if we set `VAULT_NAMESPACE`, it could end up overriding a `namespace` the user sets in their `auto_auth` config: https://github.com/hashicorp/vault/blob/9334ae5f338c90f9a6d65e749794593e50acd48d/command/agent.go#L367-L369. However, if they've explicitly set the namespace annotation, it does seem like a reasonable thing to add. Overall it's a difficult one to call, but it maybe feels safer to exclude it for now, to maintain the current behaviour?

swenson: Exactly. I only targeted the things that are in the `vault` config block.

agent-inject/agent/container_env.go
```diff
@@ -64,6 +65,48 @@ func (a *Agent) ContainerEnvVars(init bool) ([]corev1.EnvVar, error) {
...
 			Value: b64Config,
 		})
+	} else {
+		// set up environment variables to access Vault since "vault" section may not be present in the config
+		if a.Vault.Address != "" {
+			envs = append(envs, corev1.EnvVar{
+				Name:  "VAULT_ADDR",
+				Value: a.Vault.Address,
+			})
+		}
```
tvoran, line 75: (Just making sure I understand) It looks like all these can be overridden by the user with annotations from https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-annotations? So like for the vault address, the order of precedence would be:
* `vault` stanza in configmap
* This setting of VAULT_ADDR
* The value of `vault.hashicorp.com/service`

We may want to add a note about this new behavior in or adjacent to https://www.vaultproject.io/docs/platform/k8s/injector/examples

tvoran (approved):

agent-inject/agent/container_env.go
swenson: Yes. I believe that is correct. I'll double-check it before merging, and add a ticket to update those docs.

swenson: Yes, I confirmed that `vault.hashicorp.com/service` will override the agent container `VAULT_ADDR` settings.

swenson:
I noticed this also fixes another use case: even with a helm-deployed Vault, this will now set the VAULT_ADDR as expected if the `vault {}` section is missing from the config; before the agent pod would just hang forever not knowing where to connect to Vault.

swenson:
Thanks!

PR #354, opened 2022-05-24
Wait for certificate before starting HTTP listener

### LINKED ISSUE #318: Healchecks via https are succefull in case of certificate issue
I discovered strange vault injector behavior. I use liveness probe like:
```
Liveness:   http-get https://:8080/health/ready delay=30s timeout=30s period=30s #success=1 #failure=3
```
After some very strage events in may k8s cluster, vault injector failed 3 probes and was restarted by kubelet:
```
 [ERROR] handler: http: TLS handshake error from 10.0.216.118:51410: EOF
 [ERROR] handler: http: TLS handshake error from 10.0.216.118:37002: EOF
 [ERROR] handler: http: TLS handshake error from 10.0.216.118:55394: EOF
```
It is still normal and expected behavior. But after restart vault injector could not read certificate from filesystem with endless logs like:
```
handler.certwatcher: Could not load TLS keypair: tls: failed to find any PEM data in certificate input. Trying again...
handler.certwatcher: Could not load TLS keypair: tls: failed to find any PEM data in certificate input. Trying again...
...
```
I have no idea how it can be possible. I can read cert and key file inside vault injector container without problem. May be some memory issue in k8s node...
In any way, tls in vault injector is broken, but `curl https://pod_address:8080/health/ready` return 204 code as correctly working instance.
If healthcheck request fail in this scenario, kubelet will restart container and vault injector starts working normally.

May be add some try threshold for tls issue? And after threshold's excess health check endpoint will response error code.

Application deployment:

**Environment**
* Kubernetes version: 1.20.11
  * Baremetal
* vault-k8s version: 0.14.0

### LINKED ISSUE #320: Wait for handler startup until TLS has been established
**Is your feature request related to a problem? Please describe.**
Whenever the Vault Injector starts up it does a couple of things. Two of those things is to establish TLS and start up an HTTP server for serving requests. Now the requirement is there that TLS must be established before any request can come from the kubelet or kube-api to the injector's HTTP server.
Here is where the issue starts. 

Rather than waiting for the certificate to be available, either through leader election sourcing or through file sourcing, the server is started up immediately. This results in a multitude of TLS **error** messages when the readiness probe is used or any mutation requests come in. These error messages are tricky to decipher if you just start out and can be a pain for logging solutions to filter out.

**Describe the solution you'd like**
While this is expected behaviour, it is completely unnecessary. Why do we need to start up the HTTP server before any request can be served? There is no long start up time for the server and it only throws errors, and confuses users.

My solution would be to wait for the TLS to be available from either of the two sources and to then signal the main process to continue starting the HTTP handler server. This to be toggle-able by a flag, enabled by default.

The solution would not hinder the readiness check from Kubernetes since this check is unavailable to be performed correctly anyway, so the behaviour in this stays the same. Setting this value correctly is not the task of the injector but of the cluster admin.

**Describe alternatives you've considered**
None.

**Additional context**
The rest of **warning** messages related to TLS are from the cert-watcher and is a different issue. Perhaps for this a clearer warning message can be shown. Otherwise the "forcing the mutation webhook config to remain updated" janky way should be fixed:
https://github.com/hashicorp/vault-k8s/blob/286c36fcff05636581cdedcc4b2686da138ca4bc/subcommand/injector/command.go#L328

I have created a solution for this myself and would like to present this as a contribution to the project!

swenson:
When starting up, the HTTP server can start accepting requests before
the TLS config has loaded certificates from the certificate generator.

We add in a wait (assuming we don't have certificates already given to
us via flags or environment variables) for that certificate to be
generated before starting the server to prevent confusing, though
harmless, errors in logs.

* Fixes #318
* Fixes #320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 ## Unreleased
 
+Improvements:
+* ConfigMap with missing vault section should default to env vars [GH-353](https://github.com/hashicorp/vault-k8s/pull/353)
+
 Changes:
 * Certificate watcher timer deadlock fix [GH-350]([https://github.com/hashicorp/vault-k8s/pull/350)
+* Wait for certificate before starting HTTP listener [GH-354](https://github.com/hashicorp/vault-k8s/pull/354)
+* Update example injector mutating webhook config to exclude agent pod [GH-351](https://github.com/hashicorp/vault-k8s/pull/351)
 
 ## 0.16.0 (May 11, 2022)
 
--- a/helper/cert/notify.go
+++ b/helper/cert/notify.go
@@ -2,17 +2,19 @@ package cert
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
 )
 
-func NewNotify(ctx context.Context, newBundle chan<- Bundle, source Source, logger hclog.Logger) *Notify {
+func NewNotify(ctx context.Context, newBundle chan<- Bundle, notifyOnce chan<- bool, source Source, logger hclog.Logger) *Notify {
 	return &Notify{
-		ctx:    ctx,
-		ch:     newBundle,
-		source: source,
-		logger: logger,
+		ctx:        ctx,
+		ch:         newBundle,
+		notifyOnce: notifyOnce,
+		source:     source,
+		logger:     logger,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


tomhjp (approved):
LGTM

swenson:
Thanks!

Issue #310, opened 2021-12-16
Allow setting ephemeral-storage requests and limits for injected container


thechristschn:
**Is your feature request related to a problem? Please describe.**
In our cluster we deployed LimitRanges with default requests and limits for cpu, memory and ephemeral-storage.

The cpu and memory requests and limits can be configured with environment variables on the injector or annotations at the target resource. But for ephemeral-storage this isn't possible.

While our defaults in the LimitRange are good for the application containers, they are way too high for the vault-agent.
We have a multi-tenant cluster where we set LimitRanges and ResourceQuotas for all Namespaces. With the introduction of vault-agent-injector, we got the problem, that all those high defaults which are applied to the vault-agents, are depleting the quota.

Because we have many different user, we can't just change the defaults in the LimitRange to fit the vault-agent and would rather configure the values for the vault-agent container explicitly.

**Describe the solution you'd like**
Add a possibility to set ephemeral-storage requests and limits for the injected vault-agent and vault-agent-init containers. This should be possible via environment variables on the injector or via annotation on the pod.

The annotations would be `vault.hashicorp.com/agent-requests-ephemeral-storage` and `vault.hashicorp.com/agent-limits-ephemeral-storage`.

The environment variables would be `AGENT_INJECT_EPHEMERAL_STORAGE_REQUEST` and `AGENT_INJECT_EPHEMERAL_STORAGE_LIMIT`.

It should also be possible to configure the environment variables in the helm chart, but that would be a follow-up feature in the vault-helm repository.

**Describe alternatives you've considered**
We could change the defaults in the LimitRange, but that would require to adjust all other deployments in the cluster.

hsimon-hashicorp:
Thanks for submitting this request for enhancement! I'll get it triaged as soon as I can. :)

tomhjp:
Should be resolved by #360 which is now merged. Thanks!

PR #360, opened 2022-06-14
ephemeral storage resources

beorereleverion:
this PR resolve https://github.com/hashicorp/vault-k8s/issues/361 and https://github.com/hashicorp/vault-k8s/issues/310


beorereleverion:
Hello. Can anyone  review this pr?

apolubotko:
Hi, 

We have been waiting for this change for a very long time, we will be very happy to see it in the master branch

beorereleverion:
@tvoran @swenson Hello, can you review this pr? please

tomhjp (approved):
Thanks for implementing this! I gave it a test run and it all worked really nicely, with just one tiny bug where the flags specifying the defaults weren't being threaded through. I also added a little changelog note, just to save a trivial review round.

Kyslik:
This seems to be _breaking change_ :(. Can you also include a note in the docs that you have/should to specify `resources.limits.ephemeral-storage` for **all containers** in the pod otherwise you are risking to have your pods evicted more in https://github.com/kubernetes/kubernetes/blob/v1.25.0/pkg/kubelet/eviction/eviction_manager.go#L453-L479

> Although you can only specify requests and limits for individual containers, it is also useful to think about the overall resource requests and limits for a Pod. For a particular resource, a Pod resource request/limit is the sum of the resource requests/limits of that type for each container in the Pod.

From https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

---

Using ENV variable (`AGENT_INJECT_EPHEMERAL_LIMIT `) one can not disable `ephemeral-storage` limits either, because how the checks for default are implemented. The only option is to use annotation `vault.hashicorp.com/agent-limits-ephemeral` which is not ideal, because you have to update all your deployments...

tomhjp:
Thanks @Kyslik, apologies for the inconvenience. I think #386 should fix this by defaulting the ephemeral storage resources to unset, so it will then only be an explicitly opt-in feature.

Kyslik:
> Thanks @Kyslik, apologies for the inconvenience. I think #386 should fix this by defaulting the ephemeral storage resources to unset, so it will then only be an explicitly opt-in feature. Feel free to leave any comments you have in the review.

No worries; we learned something new here :). Thank you.

PR #363, opened 2022-06-16
Set request UID when returning admission errors

tvoran:
Otherwise the error returned from vault-k8s will not show up in the events of the object that owns the Pod (ex. a ReplicaSet). I suspect the UID started being required with v1 of the admission API.

Currently without this patch, kube-apiserver logs something like this when vault-k8s returns an AdmissionError:

```
E0616 23:21:07.674311       1 dispatcher.go:184] failed calling webhook "vault.hashicorp.com": received invalid webhook response: expected response.uid="3d19f57a-2a0e-4bc9-8a0e-d54c413b21e5", got ""
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+Bugs:
+* Properly return admission errors [GH-363](https://github.com/hashicorp/vault-k8s/pull/363)
+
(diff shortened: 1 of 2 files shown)


tomhjp (approved):

swenson (approved):
LGTM

PR #371, opened 2022-07-18
go and dep updates

tvoran:
Updated go to 1.17.12, and removed the `upgrade libretls` line from Dockerfile since it's no longer necessary (added in #327). Updated `golang.org/x/net` to avoid CVE-2021-44716 for hygiene sake and to reduce vulnerability scanner noise.

Steps to update `golang.org/x/net`:
```
g set 1.17.12
go get -u golang.org/x/net
go mod tidy
```

With these changes, a trivy scan comes back with no vulnerabilities:
```
make image
trivy image hashicorp/vault-k8s:0.0.0-dev
```


Dockerfile
```diff
@@ -38,7 +38,6 @@ RUN addgroup vault && \
 
 # Set up certificates, base tools, and software.
 RUN set -eux && \
-    apk update && apk upgrade libretls && \
     apk add --no-cache ca-certificates libcap su-exec iputils
```
tomhjp, line 41: Based on your PR description, I guess it still works, but should we keep the `apk update` command to ensure we get the latest for these packages?

tomhjp (approved):

mickael-hc (commented):
👍 the base image is tracking Alpine 3.16.1, I think it's fine to remove the call to `apk upgrade` here

Dockerfile
tvoran: I feel like that could make the build slightly less reproducible? Though since this Dockerfile is referencing "alpine:latest" instead of a specific version, reproducible is probably a moot point here, so I'll keep the update call.

(For context, the whole line was introduced just to fix the libretls issue here: https://github.com/hashicorp/vault-k8s/pull/327/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R41)

PR #372, opened 2022-07-22
Run acceptance tests from vault-helm repo

tomhjp:
Run the acceptance tests from the vault-helm repo using a matrix testing strategy to ensure basic compatibility with all recent Kubernetes minor versions.

I wanted to try to re-use the helm tests instead of having to maintain an additional set of complex acceptance tests, but lmk if you think this repo should have its own test suite instead. It could also be that we just use the helm tests as a base set and add more bats tests within this repo in the future to cover injector specific things like mutating web hook API versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -18,11 +18,11 @@ jobs:
           echo "::set-output name=go-version::$(cat .go-version)"
 
   test:
+    env:
+      TARBALL_FILE: vault-k8s-image.docker.tar
     needs:
       - get-go-version
     runs-on: ubuntu-latest
-    env:
-      GO111MODULE: on
     steps:
       - uses: actions/checkout@v2
 
@@ -41,10 +41,17 @@ jobs:
           go-version: ${{ needs.get-go-version.outputs.go-version }}
 
       - name: Build
-        run: make build
+        run: |
+          REGISTRY=hashicorp make build image
+          docker save --output "${TARBALL_FILE}" hashicorp/vault-k8s:0.0.0-dev
 
       - name: Test
         run: make test
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: vault-k8s-image
+          path: ${{ env.TARBALL_FILE }}
 
   acceptance:
     needs:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -45,3 +45,63 @@ jobs:
...
+      # Checkout vault-helm for acceptance test code.
+      - uses: actions/checkout@v2
+        with:
+          repository: "hashicorp/vault-helm"
+          ref: "v0.20.1"
+          path: "vault-helm"
+
+      - run: |
+          yq --in-place --yaml-roundtrip '.injector.image.tag |= "0.0.0-dev"' ./vault-helm/values.yaml
+          CLEANUP=false bats ./vault-helm/test/acceptance -t --filter injector
```
tvoran, line 107: I'm curious why `CLEANUP=false` here?

.github/workflows/tests.yaml
```diff
@@ -45,3 +45,63 @@ jobs:
...
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - run: |
+          REGISTRY=hashicorp make -C vault-k8s image
```
benashz, line 95: nice 👍

test/kind/config.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Cluster
```
benashz, line 1: Not sure if it makes sense, but might be able to use the kind config from vault-helm: https://github.com/hashicorp/vault-helm/blob/main/test/kind/config.yaml

It does have some extra bits, so not sure.

tvoran (approved):
This is great! I like the idea of keeping at least the base integration tests in one spot.

.github/workflows/tests.yaml
```diff
@@ -45,3 +45,63 @@ jobs:
...
+
+      # Checkout vault-helm for acceptance test code.
+      - uses: actions/checkout@v2
+        with:
+          repository: "hashicorp/vault-helm"
+          ref: "v0.20.1"
+          path: "vault-helm"
+
+      - run: |
+          yq --in-place --yaml-roundtrip '.injector.image.tag |= "0.0.0-dev"' ./vault-helm/values.yaml
```
benashz, line 106: nit: at some point it would be good to be able to pass extra `helm` arguments to the bats tests, rather than munging `values.yaml`. This would have the added benefit of not needing the extra step of installing `yq`

benashz (approved):
Added a few suggestions, otherwise looks great!

test/kind/config.yaml
tomhjp: Good point, the extra stuff is harmless and if it needs to diverge in the future, we can create a special config for vault-k8s kind at that point.

.github/workflows/tests.yaml
tomhjp: Oops, this is some hangover from some debugging output I used to confirm that I really was testing the tag I thought I was.

.github/workflows/tests.yaml
tomhjp: Yeah agreed 👍

PR #393, opened 2022-10-12
update golang.org/x/text to v0.3.8

tvoran:
To avoid [GO-2022-1059](https://pkg.go.dev/vuln/GO-2022-1059)

```
go get golang.org/x/text@latest
go mod tidy
```


swenson (approved):
LGTM

PR #395, opened 2022-10-19
Support building ubi images in CI

mdeggies:


.github/workflows/build.yml
```diff
@@ -20,33 +20,34 @@ jobs:
...
       go-version: ${{ steps.get-go-version.outputs.go-version }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Determine Go version
         id: get-go-version
         run: |
           echo "Building with Go $(cat .go-version)"
-          echo "::set-output name=go-version::$(cat .go-version)"
+          echo "go-version=$(cat .go-version)" >> $GITHUB_OUTPUT
```
mdeggies, line 28: This is just some cleanup that gets rid of those set-output errors in the actions UI

.github/workflows/build.yml
```diff
@@ -20,33 +20,34 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: get product version
         id: get-product-version
         run: |
           VERSION="${{ github.event.inputs.version || '0.0.0-dev' }}"
           echo "Using version ${VERSION}"
-          echo "::set-output name=product-version::${VERSION}"
+          echo "product-version=${VERSION}" >> $GITHUB_OUTPUT
```
mdeggies, line 41: This is just some cleanup that gets rid of a couple set-output errors in the actions UI

.github/workflows/build.yml
```diff
@@ -117,3 +118,24 @@ jobs:
...
+      version: ${{needs.get-product-version.outputs.product-version}}
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: hashicorp/actions-docker-build@v1
+        with:
+          version: ${{env.version}}
+          target: ubi
+          arch: amd64
+          redhat_tag: scan.connect.redhat.com/ospid-60f9fdbec3a80eac643abedf/${{env.repo}}:${{env.version}}-ubi
```
mdeggies, line 139: As commented, this tag will need to change to `quay.io/redhat-isv-containers/5f89bbc30b94cf64cfeb500b/${{env.repo}}:${{env.version}}-ubi` on October 20th

Dockerfile
```diff
@@ -48,6 +59,59 @@ COPY dist/$TARGETOS/$TARGETARCH/vault-k8s /bin/
...
+ENV PRODUCT_VERSION=$PRODUCT_VERSION
+
+ARG PRODUCT_NAME=$BIN_NAME
+
+# TARGETOS and TARGETARCH are set automatically when --platform is provided.
+ARG TARGETOS TARGETARCH
+
+# Set ARGs as ENV so that they can be used in ENTRYPOINT/CMD
+ENV PRODUCT_VERSION=$PRODUCT_VERSION
+ENV BIN_NAME=$BIN_NAME
```
tvoran, line 82: It looks like these were already defined on lines 72-73?

tvoran (commented):
This passed the openshift preflight for me, and our vault-helm injector acceptance tests. I think the ospid isn't quite right (but of course that tag should be changing tomorrow anyway).

I think we'll probably want to remove the `prod-ubi-image` target in the Makefile as well.

.github/workflows/build.yml
samsalisbury: Red Hat have made their changes live, the tag format will now be as follows (they've dropped the product name from the tags entirely, and use the project ID not the old `ospid-...`):

```suggestion
          redhat_tag: quay.io/redhat-isv-containers/5f89bbc30b94cf64cfeb500b:${{env.version}}-ubi
```

cc @mdeggies

mdeggies: Thanks, Sam! I committed your suggestion.

Dockerfile
mdeggies: Oh good catch, ty!

mdeggies:
Thanks for the review! I removed the make target and removed those duplicates lines in the dockerfile. Lmk if you notice anything else -- I confirmed the project ID in redhat, and the necessary secrets (registry key + project ID) have been set in vault.

tvoran (approved):

PR #398, opened 2022-10-24
Switch to newer json-patch; remove dead code

swenson:
Switch from `github.com/mattbaird/jsonpatch` (unmaintained) to a new version of `github.com/evanphx/json-patch`.

I also noticed that the `agent.Patches` object is unused, as is the `patches` parameter of `agent.New()`, so I removed those, which simplifies some of the code.

(This is in preparation for some more work involving JSON patches, so I wanted to clean up existing usage of it.)

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -7,12 +7,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/evanphx/json-patch"
 	"github.com/hashicorp/vault/sdk/helper/strutil"
-	"github.com/mattbaird/jsonpatch"
 	corev1 "k8s.io/api/core/v1"
 )
-
-// TODO swap out 'github.com/mattbaird/jsonpatch' for 'github.com/evanphx/json-patch'
 
 const (
 	DefaultVaultImage                       = "hashicorp/vault:1.12.0"
@@ -72,9 +70,6 @@ type Agent struct {
 
 	// Namespace is the Kubernetes namespace the request originated from.
 	Namespace string
-
-	// Patches are all the mutations we will make to the pod request.
-	Patches []*jsonpatch.JsonPatchOperation
 
 	// Pod is the original Kubernetes pod spec.
 	Pod *corev1.Pod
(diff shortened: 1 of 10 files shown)


agent-inject/handler_test.go
```diff
@@ -601,13 +401,16 @@ func TestHandlerHandle(t *testing.T) {
...
 			if len(resp.Patch) > 0 {
 				req.NoError(json.Unmarshal(resp.Patch, &actual))
 				for i := range actual {
-					actual[i].Value = nil
+					delete(actual[i], "value")
 				}
 			}
+			for i := range tt.Patches {
+				delete(tt.Patches[i], "value")
+			}
```
tvoran, line 413: Was this extra `delete` loop for the expected patches necessary with the new jsonpatch, or just being prudent?

agent-inject/agent/agent.go
```diff
@@ -524,49 +518,47 @@ func ShouldInject(pod *corev1.Pod) (bool, error) {
 
 // Patch creates the necessary pod patches to inject the Vault Agent
 // containers.
-func (a *Agent) Patch() ([]byte, error) {
-	var patches []byte
-
+func (a *Agent) Patch() (patches jsonpatch.Patch, patchesBytes []byte, err error) {
```
tvoran, line 521: Is this returning `patches jsonpatch.Patch` just for use in tests? I'm not seeing that parameter used in the actual code. I'm wondering if it needs to be returned, and patchesBytes could just be unmarshal'ed in the test.

tvoran (commented):
This looks pretty great; love the internal json patch helpers. Just a couple style comments so far.

agent-inject/agent/agent.go
```diff
@@ -584,9 +576,10 @@ func (a *Agent) Patch() ([]byte, error) {
 
 	// Init Container
 	if a.PrePopulate {
-		container, err := a.ContainerInitSidecar()
+		var container corev1.Container
+		container, err = a.ContainerInitSidecar()
 		if err != nil {
-			return patches, err
+			return
```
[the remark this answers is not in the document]
swenson: Yeah, in general I agree. I'll see if I can rework this.

agent-inject/handler_test.go
swenson: It's necessary now to remove the value field from the test cases. I tried passing in `nil`, but it still generates some bytes (JSON `null`) I think, so the two won't be equal below.

tomhjp (approved):
LGTM!

tvoran (approved):

swenson:
Thanks!

PR #399, opened 2022-10-26
Allow modifying agent and agent-init containers

swenson:
Adds the annotations:

* `vault.hashicorp.com/agent-init-json-patch` and
* `vault.hashicorp.com/agent-json-patch`

To a pod to allow the pod to specify arbitrary modifications to the `vault-agent` and `vault-agent-init` containers that are injected into the pod.

Here an example patch I applied to a deployment:

```yaml
spec:
  template:
    metadata:
      annotations:
        vault.hashicorp.com/role: "internal-app"
        vault.hashicorp.com/log-level: "debug"
        vault.hashicorp.com/agent-inject-secret-credentials.txt: 'secret/data/somesecret/config'
        vault.hashicorp.com/agent-json-patch: '[{"op": "replace", "path": "/name", "value": "different-name"}]'
```

And verified that the name is indeed changed:

```sh
$ kubectl describe pod orgchart-6f9789489c-vwzb6

Containers:
...
  different-name:
    Container ID:  containerd://93a4207a7f78fc424cd829b5430908a91133a49521e7f19b468fc57dbdcfd94c
    Image:         hashicorp/vault:1.11.3
...
```

(This also updates to Go 1.19 to make tests slightly nicer.)


agent-inject/agent/annotations.go
```diff
@@ -526,6 +535,30 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
+		// ignore empty string
+		if pod.ObjectMeta.Annotations[AnnotationAgentJsonPatch] == "" {
+			delete(pod.ObjectMeta.Annotations, AnnotationAgentJsonPatch)
+		} else {
+			_, err := jsonpatch.DecodePatch([]byte(pod.ObjectMeta.Annotations[AnnotationAgentJsonPatch]))
+			if err != nil {
+				return fmt.Errorf("error parsing JSON patch for annotation %s: %+v", AnnotationAgentJsonPatch, err)
+			}
+		}
+	}
```
tomhjp, line 549: Might as well to avoid repetition and possible errors on future edits?

```suggestion
	if patch, ok := pod.ObjectMeta.Annotations[AnnotationAgentJsonPatch]; ok {
		// ignore empty string
		if patch == "" {
			delete(pod.ObjectMeta.Annotations, AnnotationAgentJsonPatch)
		} else {
			_, err := jsonpatch.DecodePatch([]byte(patch))
			if err != nil {
				return fmt.Errorf("error parsing JSON patch for annotation %s: %+v", AnnotationAgentJsonPatch, err)
			}
		}
	}
```

agent-inject/agent/container_init_sidecar.go
```diff
@@ -91,5 +93,27 @@ func (a *Agent) ContainerInitSidecar() (corev1.Container, error) {
...
+	}
+	patch, err := jsonpatch.DecodePatch([]byte(a.InitJsonPatch))
+	if err != nil {
+		return newContainer, fmt.Errorf("failed to decode JSON patch: %v", err)
+	}
+	newContainerJson, err := patch.Apply(containerJson)
+	if err != nil {
+		return newContainer, fmt.Errorf("failed to apply JSON patch: %+v", err)
+	}
+	newNewContainer := corev1.Container{}
```
tomhjp, line 113: 😂 A+ variable name

agent-inject/agent/annotations.go
```diff
@@ -526,6 +535,30 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
 
+	// validate JSON patches
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationAgentJsonPatch]; ok {
+		// ignore empty string
+		if pod.ObjectMeta.Annotations[AnnotationAgentJsonPatch] == "" {
+			delete(pod.ObjectMeta.Annotations, AnnotationAgentJsonPatch)
+		} else {
+			_, err := jsonpatch.DecodePatch([]byte(pod.ObjectMeta.Annotations[AnnotationAgentJsonPatch]))
+			if err != nil {
+				return fmt.Errorf("error parsing JSON patch for annotation %s: %+v", AnnotationAgentJsonPatch, err)
```
tomhjp, line 546: In a few places in this PR you've used `%v` or `%+v` for errors, but I think `%s` is more common when printing, and `%w` is preferred when wrapping (as in this case) - any particular reason for `%+v`?

tomhjp (approved):
Nice! LGTM. It would be nice to use `%w` for wrapping errors, but nothing else of substance.

agent-inject/agent/container_init_sidecar.go
swenson: lol. I meant to change that. :)

agent-inject/agent/container_sidecar_test.go
```diff
@@ -1275,3 +1276,220 @@ func TestContainerCache(t *testing.T) {
...
+			},
+			RunAsGroup:               optional[int64](100),
+			RunAsUser:                optional[int64](1000),
+			RunAsNonRoot:             optional[bool](true),
+			ReadOnlyRootFilesystem:   optional[bool](true),
+			AllowPrivilegeEscalation: optional[bool](false),
+		},
+	}
+
+	baseInitContainer := corev1.Container{
```
tvoran, line 1320: Since this is almost identical to baseContainer, we could maybe make a copy and then change the fields that are different, instead of redefining it all.

agent-inject/agent/container_init_sidecar.go
```diff
@@ -91,5 +93,27 @@ func (a *Agent) ContainerInitSidecar() (corev1.Container, error) {
...
+	}
+	newContainerJson, err := patch.Apply(containerJson)
+	if err != nil {
+		return newContainer, fmt.Errorf("failed to apply JSON patch: %w", err)
+	}
+	newContainer = corev1.Container{}
+	err = json.Unmarshal(newContainerJson, &newContainer)
+	if err != nil {
+		return newContainer, err
+	}
```
tvoran, line 117: The sloshing back and forth here between marshal'ing and unmarshal'ing the container gives me a little pause, but I don't see a way around it. (Not a big deal either, just thinking out loud.)

tvoran (approved):
This is a really great idea!

jasonodonnell (approved):
LGTM!

agent-inject/agent/container_init_sidecar.go
swenson: Yeah, same. We'd have to have something like a StructurePatch or MapStructurePatch to combine JSON Patch + modifying Go structures directly, but that is definitely outside the scope of this PR.

agent-inject/agent/container_sidecar_test.go
swenson: :+1:

swenson:
Thanks!

PR #400, opened 2022-10-28
Add exit on error auto auth annotation

swenson:
Support was added in the Vault agent in https://github.com/hashicorp/vault/pull/17091

Additionally, I tested this on a cluster by adding the `vault.hashicorp.com/agent-auto-auth-exit-on-err: "true"` annotation on a pod, and observing that the agent configuration was:

```json
{
  "auto_auth": {
    "method": {
      "type": "kubernetes",
      "mount_path": "auth/kubernetes",
      "config": {
        "role": "internal-app",
        "token_path": "/var/run/secrets/kubernetes.io/serviceaccount/token"
      },
      "exit_on_err": true
    }...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+Features:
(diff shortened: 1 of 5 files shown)


tvoran (approved):

tvoran:
Oh and don't forget to add a changelog entry as well.

swenson:
Thanks!

PR #401, opened 2022-11-15
fix: update cert secret if exists

### LINKED ISSUE #378: vault-injector-certs Secret metadata is regularly overwritten, breaking any gitops flow
**Describe the bug**
During Agent Sidecar Injector installation via helm, Secret `vault-injector-certs` is created empty:
```
apiVersion: v1
kind: Secret
metadata:
  name: vault-injector-certs
  namespace: vault-k8s
  labels:
    app.kubernetes.io/name: vault-agent-injector
    app.kubernetes.io/instance: vault
    app.kubernetes.io/managed-by: Helm
```

Then, Secret is filled with data (certificate) probably by some controller but unfortunately this process is removing metadata (labels and annotations) from the Secret:
```
apiVersion: v1
kind: Secret
metadata:
  name: vault-injector-certs
  namespace: vault-k8s
data:
  cert: <redacted>
  key: <redacted>
type: Opaque
```

So any gitops based flow (e.g. with ArgoCD) is broken due to reconciliation loop constantly detecting changes that need to be sync with the source of true (helm).
Additionally by removing annotations and labels all object tracking info is wiped out from the object (ArgoCD is using labels or annotation to keep track of managed objects)

**To Reproduce**
1. Deploy vault-agent injection via helm
2. Wait until Secret `vault-injector-certs` is filled with data
3. Secret metadata (labels and annotations) is removed.

**Expected behavior**
Controller or other mechanism is not removing metadata from Secret `vault-agent injection` during filling it with data.

**Environment**
* Kubernetes version:
```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:21:56Z", GoVersion:"go1.18.4", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:23:26Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
```
* vault-k8s version: 0.16.1

busser:
Fixes #378.


busser:
@tvoran no rush, but if you have time to take a look that would be great 🙂

helper/cert/source_gen.go
```diff
@@ -236,25 +236,31 @@ func (s *GenSource) retryUpdateSecret(ctx context.Context, leaderCh chan bool, b
...
-		},
-		Data: map[string][]byte{
-			"cert": bundle.Cert,
-			"key":  bundle.Key,
-		},
-	}
-	// Attempt updating the Secret first, and if it doesn't exist, fallback to
-	// create
-	_, err := s.K8sClient.CoreV1().Secrets(s.Namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	secret, err := s.K8sClient.CoreV1().Secrets(s.Namespace).Get(ctx, certSecretName, metav1.GetOptions{})
```
tvoran, line 239: I think we may want to do the get from the informer cache here, instead of the regular k8s client. That should help avoid any rate limiting from the k8s API, even though this get shouldn't be firing that often. Something like:

```suggestion
	secret, err := s.SecretsCache.Lister().Secrets(s.Namespace).Get(certSecretName)
```

Similar to getBundleFromSecret(): https://github.com/hashicorp/vault-k8s/blob/0603c2a03300773fd6ba6174c6d4327d70889d58/helper/cert/source_gen.go#L266-L269

busser: It's done 👍

tvoran:
Looks like we'll also need to update the tests, since `make test` crashes in `TestGenSource_leader()`. I bet the secret informer will need to be setup in that test, similar to `TestGenSource_follower()`.

It would also be great to add another test that starts with an empty secret with labels and annotations, runs the cert generation, then checks that the updated secret's labels and annotations are intact. Maybe another variant of `TestGenSource_leader()`?

busser:
`TestGenSource_leader()` now fixed. Like you said, it just needed the informer to be setup.

I added a `TestGenSource_leader_expiry()`, which combines logic from `TestGenSource_leader` and `TestGenSource_expiry` to confirm that the secret retains its metadata when it is updated with a new bundle.

tvoran (approved):
Thanks for adding the tests!

kenske:
Great work, thanks! Any idea when this will get released?

tvoran:
@kenske Sometime later this month is the best estimate I can give right now.

PR #407, opened 2022-12-15
Patch TLS secret instead of update to keep labels

claudex:
Use a patch method instead of update for the injector secrets to avoid an override of the labels in the secrets and prevent a conflict between the helm chart which add the labels and the injector which remove them.


claudex:
I didn't see #401 was already there. So this is not needed anymore.

PR #411, opened 2022-12-23
Set Kubernetes user-agent to include version

swenson:
We weren't setting the user-agent, so Kubernetes sets a default of the command-line (`vault-k8s`, which is okay) and an automatically set version in the Kubernetes code base. This is probably not what we want, since it implies that the Kubernetes version is the `vault-k8s` version, and it is often just `0.0.0`. See https://github.com/kubernetes/client-go/blob/master/pkg/version/base.go

Instead, we set the user-agent based on our own version.

I didn't see an easy way to test this in a unit test, but I did test it manually by running `vault-k8s` with the environment variable `GODEBUG=http2debug=1` set.

### DIFF
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -22,6 +22,7 @@ import (
 	agentInject "github.com/hashicorp/vault-k8s/agent-inject"
 	"github.com/hashicorp/vault-k8s/helper/cert"
 	"github.com/hashicorp/vault-k8s/leader"
+	"github.com/hashicorp/vault-k8s/version"
 	"github.com/mitchellh/cli"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	adminv1 "k8s.io/api/admissionregistration/v1"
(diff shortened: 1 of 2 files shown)


subcommand/injector/command.go
```diff
@@ -117,6 +118,7 @@ func (c *Command) Run(args []string) int {
 		c.UI.Error(fmt.Sprintf("Error loading in-cluster K8S config: %s", err))
 		return 1
 	}
+	config.UserAgent = fmt.Sprintf("vault-k8s/%s", version.Version)
```
tvoran, line 121: Should this include the "v" prefix?

```suggestion
	config.UserAgent = fmt.Sprintf("vault-k8s/v%s", version.Version)
```

swenson: Nope. It's not obvious, but the `v` is added by the release process here: https://github.com/hashicorp/vault-k8s/blob/main/Makefile#L14

tvoran: Ah! Gotcha, thanks.

tvoran (approved):

tvoran:
Probably should update the changelog here too.

swenson:
Thanks!

PR #413, opened 2023-01-11
VAULT-5280 add telemetry annotations for agent sidecar

kschoche:
* Adds a new annotation for `vault.hashicorp.com/agent-telemetry` which defines the [telemetry](https://developer.hashicorp.com/vault/docs/configuration/telemetry) configuration for the agent sidecar. All parameters linked above are currently supported.
* The annotation behaves similarly to [vault.hashicorp.com/auth-config](https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations#vault-hashicorp-com-auth-config) in that you specify an annotation via `vault.hashicorp.com/agent-telemetry-<key>: "<value>"`

An example which configures prometheus telemetry on the agent sidecar for an application:
```
annotations:
	"vault.hashicorp.com/agent-telemetry-prometheus_retention_time": "30s",
	"vault.hashicorp.com/agent-telemetry-disable_hostname":          "true",
```
..which results in the following configuration stanza being added to the sidecar environment via rendering the environment:
```
telemetry {
  prometheus_retention_time = "30s"
  disable_hostname = true
}
```

Note to Reviewers: docs PR https://github.com/hashicorp/vault/pull/18681


agent-inject/agent/annotations.go
```diff
@@ -759,8 +764,8 @@ func (a *Agent) setShareProcessNamespace(pod *corev1.Pod) (bool, error) {
 	if pod.Spec.ShareProcessNamespace != nil {
 		if !*pod.Spec.ShareProcessNamespace && shareProcessNamespace {
 			return DefaultAgentShareProcessNamespace,
-			errors.New("shareProcessNamespace explicitly disabled on the pod, " +
-				"refusing to enable it")
+				errors.New("shareProcessNamespace explicitly disabled on the pod, " +
```
kschoche, line 767: This was a lint cleanup, I can revert it if necessary.

agent-inject/agent/config.go
```diff
@@ -119,6 +120,55 @@ type TemplateConfig struct {
...
+// Telemetry defines the configuration for telemetry in Vault Agent.
+type Telemetry struct {
+	UsageGaugePeriod                   string `json:"usage_gauge_period,omitempty"`
+	MaximumGaugeCardinality            int    `json:"maximum_gauge_cardinality,string,omitempty"`
+	DisableHostname                    bool   `json:"disable_hostname,string,omitempty"`
+	EnableHostnameLabel                bool   `json:"enable_hostname_label,string,omitempty"`
+	LeaseMetricsEpsilon                string `json:"lease_metrics_epsilon,omitempty"`
+	AddLeaseMetricsNamespaceLabels     bool   `json:"add_lease_metrics_namespace_labels,string,omitempty"`
+	FilterDefault                      bool   `json:"filter_default,string,omitempty"`
+	PrefixFilter                       string `json:"prefix_filter,omitempty"`
```
kschoche, line 132: The `prefix_filter` is actually a `[]string` but I could not figure out how to get `json.Unmarshal()` to parse it properly, I *think* this still works though, would like to validate it though.

kschoche: This was resolved by doing a `json.Unmarshal()` directly into an interface type when we are accumulating the telemetry annotations, thanks Ben!

agent-inject/agent/annotations_test.go
```diff
@@ -1243,3 +1243,46 @@ func TestDisableKeepAlives(t *testing.T) {
...
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			pod := testPod(tc.annotations)
+			agentConfig := basicAgentConfig()
+			err := Init(pod, agentConfig)
+			require.NoError(t, err)
+			agent, err := New(pod)
+			require.NoError(t, err)
+			require.Equal(t, true, reflect.DeepEqual(tc.expectedValues, agent.Vault.AgentTelemetryConfig))
```
kschoche, line 1285: I don't normally use this pattern for `require.Equal()` because it obscures the diffs that are dumped on failures, but in this case it's functionally equivalent to:
```
			require.Equal(t, tc.expectedValues, agent.Vault.AgentTelemetryConfig)
```

agent-inject/agent/annotations_test.go
```diff
@@ -1243,3 +1243,46 @@ func TestDisableKeepAlives(t *testing.T) {
...
 }
+
+func TestParseTelemetryAnnotations(t *testing.T) {
+	tests := map[string]struct {
+		annotations    map[string]string
+		expectedValues map[string]interface{}
+	}{
+		"prometheus": {
+			annotations: map[string]string{
+				"vault.hashicorp.com/agent-telemetry-prometheus_retention_time": "5s",
```
swenson, line 1254: I am surprised that the `json.Unmarshal` above works on this: `5s` isn't a valid JSON string (thought `"5s"` is).

I feel like we would need to have an escape hatch to treat the parameter as a string in the default case.

But that would presume that this test is failing, which it isn't?

I'm confused :)

kschoche: I *think* it's working the way it should? Maybe I'm not following why we would need the escape hatch?

swenson: So, I guess I don't understand how this is working, honestly:

* The annotation is `5s`. Not `"5s"`.
* This gets passed as the `value` into: https://github.com/hashicorp/vault-k8s/pull/413/files#diff-8e04ce2f98b3799b387789c744675d1f7d5ad9751312f2deee564b461b9384eaR854
* That should return an error:

https://go.dev/play/p/ihsDQQbuO_h

swenson: Okay, I am mistaken -- there already is an escape hatch. I was misreading the code. The `if` statement above is only triggered if `err != nil`, i.e., there was a JSON parsing error. I had everything kind of backwards.

swenson (approved):
LGTM

agent-inject/agent/annotations_test.go
kschoche: Annotations in kubernetes are always `map[string]string{}` so it will always get passed in as a string right? However, `[]byte{'5s'}` is not a string?

kschoche: Ahhhh, okay now I see why there was confusion! Sorry about that :D

swenson: No worries, it was purely confusion on my end. :)

agent-inject/agent/config_test.go
```diff
@@ -766,3 +766,58 @@ func TestConfigAgentQuit(t *testing.T) {
...
+	}{
+		{
+			"annotations that exercise all of the types",
+			map[string]string{
+				"vault.hashicorp.com/agent-telemetry-prometheus_retention_time": "5s",
+				"vault.hashicorp.com/agent-telemetry-prefix_filter":             "[\"+vault.token\", \"-vault.expire\", \"+vault.expire.num_leases\"]",
+				"vault.hashicorp.com/agent-telemetry-maximum_gauge_cardinality": "3",
+				"vault.hashicorp.com/agent-telemetry-lease_metrics_epsilon":     "foo",
+				"vault.hashicorp.com/agent-telemetry-enable_hostname_label":     "true",
+				"vault.hashicorp.com/agent-telemetry-disable_hostname":          "true",
```
benashz, line 784: For completeness, would it make it sense to include all possible annotations here?

benashz (commented):
Looking good! I only had one suggestion for the tests.

agent-inject/agent/config_test.go
kschoche: Yeah! I can definitely add that.

agent-inject/agent/config_test.go
```diff
@@ -766,3 +766,108 @@ func TestConfigAgentQuit(t *testing.T) {
...
+				"vault.hashicorp.com/agent-telemetry-circonus_check_id":                      "foo",
+				"vault.hashicorp.com/agent-telemetry-circonus_check_force_metric_activation": "false",
+				"vault.hashicorp.com/agent-telemetry-circonus_check_instance_id":             "foo:bar",
+				"vault.hashicorp.com/agent-telemetry-circonus_check_search_tag":              "foo:bar",
+				"vault.hashicorp.com/agent-telemetry-circonus_check_display_name":            "foo",
+				"vault.hashicorp.com/agent-telemetry-circonus_check_tags":                    "foo,bar",
+				"vault.hashicorp.com/agent-telemetry-circonus_broker_id":                     "foo",
+				"vault.hashicorp.com/agent-telemetry-circonus_broker_select_tag":             "foo:bar",
+				"vault.hashicorp.com/agent-telemetry-dogstatsd_addr":                         "https://foo.com",
+				"vault.hashicorp.com/agent-telemetry-dogstatsd_tags":                         "[\"foo:bar\", \"foo:baz\"]",
```
benashz, line 803: nit: we typically prefer using back quotes instead of escaping the `"`, whenever possible. 

```suggestion
				"vault.hashicorp.com/agent-telemetry-dogstatsd_tags":                         `["foo:bar", "foo:baz"]`,
```

agent-inject/agent/config_test.go
```diff
@@ -766,3 +766,108 @@ func TestConfigAgentQuit(t *testing.T) {
...
+				"vault.hashicorp.com/agent-telemetry-prometheus_retention_time":              "24h",
+				"vault.hashicorp.com/agent-telemetry-stackdriver_project_id":                 "foo",
+				"vault.hashicorp.com/agent-telemetry-stackdriver_location":                   "useast-1",
+				"vault.hashicorp.com/agent-telemetry-stackdriver_namespace":                  "foo",
+				"vault.hashicorp.com/agent-telemetry-stackdriver_debug_logs":                 "false",
+				"vault.hashicorp.com/agent-telemetry-prefix_filter":                          "[\"+vault.token\", \"-vault.expire\", \"+vault.expire.num_leases\"]",
```
benashz, line 809:
```suggestion
				"vault.hashicorp.com/agent-telemetry-prefix_filter":                          `["+vault.token", "-vault.expire", "+vault.expire.num_leases"]`, 
```

benashz (approved):
Just a few minor nits to address, then :+1:

Issue #377, opened 2022-08-03
Add ability to run Vault Agent Sidecar in the namespace compliant with restricted Pod Security Standard


r0bj:
**Is your feature request related to a problem? Please describe.**
It's becoming more common to restrict application namespaces with Pod Security Admission (https://kubernetes.io/docs/concepts/security/pod-security-admission/) especially with restricted Pod Security Standard (https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted).
Unfortunately currently `securityContext` in `vault-agent-init` init container is configured in a way that it's not possible to run it in a namespace compliant with restricted Pod Security Standard. It's because there is a missing field `seccompProfile`:
```
  Warning  FailedCreate  22s               replicaset-controller  Error creating: pods "nginx-test-867b89cf7-8jhlj" is forbidden: violates PodSecurity "restricted:latest": seccompProfile (pod or container "vault-agent-init" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

**Describe the solution you'd like**
`securityContext` for `vault-agent-init` currently looks similar to:
```
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      readOnlyRootFilesystem: true
      runAsGroup: 1000
      runAsNonRoot: true
      runAsUser: 100
```

So we should probably add:
```
      seccompProfile:
        type: RuntimeDefault
```

in result `securityContext` would look like this:
```
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      readOnlyRootFilesystem: true
      runAsGroup: 1000
      runAsNonRoot: true
      runAsUser: 100
      seccompProfile:
        type: RuntimeDefault
```

This change would allow to run Vault Agent Sidecar in namespace compliant with restricted Pod Security Standard.

andrejvanderzee:
Is there a workaround to set `seccompProfile.type` to `RuntimeDefault`?

tvoran:
Hi folks, you should be able to use the json patch annotation to set the seccompProfile type if you like:

```yaml
vault.hashicorp.com/agent-json-patch: '[{"op": "replace", "path": "/securityContext/seccompProfile", "value": {"type": "RuntimeDefault"}}]'
```

If I recall correctly, the seccompProfile at the Pod level should be inherited by the containers, so I wonder if you've tried setting it at the Pod level?

andrejvanderzee:
[...]

PR #416, opened 2023-01-18
Add SeccompProfile to SecurityContext

andrejvanderzee:
The generated sidecar violates the PodSecurity `restricted` and therefore is not admitted:

``` 
Warning  FailedCreate  3m23s (x8 over 8m48s)  replicaset-controller  (combined from similar events): Error creating: pods "perfana-grafana-79b685674b-rgfsf" is forbidden: violates PodSecurity "restricted:latest": seccompProfile (pod or containers "vault-agent-init", "vault-agent" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

As a result, we have to lower our security standard to `pod-security.kubernetes.io/enforce:  baseline`.

This pull request adds `SeccompProfile` to the generates `secuirtContext` for the sidecar.

Also noted in https://github.com/hashicorp/vault-k8s/issues/377.

### DIFF
--- a/agent-inject/agent/\
+++ b/agent-inject/agent/\
@@ -1336,9 +1336,6 @@ func TestAgentJsonPatch(t *testing.T) {
 			RunAsNonRoot:             optional[bool](true),
 			ReadOnlyRootFilesystem:   optional[bool](true),
 			AllowPrivilegeEscalation: optional[bool](false),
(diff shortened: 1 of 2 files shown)


tvoran:
Closing for now (see discussion on #377).

PR #422, opened 2023-02-22
ci: use go-version-file and pin actions

tvoran:
Uses the `go-version-file` option in the setup-go action, instead of a separate job for getting the version from .go-version. Pins all github actions to SHAs (and upgrades a few). Bumps go to 1.19.6.


.github/workflows/build.yml
```diff
@@ -14,25 +14,12 @@ env:
   PKG_NAME: "vault-k8s"
 
 jobs:
-  get-go-version:
```
tomhjp, line 17: Nice!

tomhjp (approved):
LGTM. Should we maybe add a dependabot config for github actions as well to help keep these up to date now that they're pinned?

swenson (approved):
LGTM

tvoran:
@tomhjp Good idea, will add one in a follow-on PR.

PR #435, opened 2023-02-28
dockerfile: explicit alpine versions

tvoran:
Made the alpine image versions explicit so dependabot can parse them for updates. Also increased the docker update frequency to daily, to pick up UBI image updates in particular, since security issues with them will not be caught until they're submitted to redhat.


swenson (approved):
LGTM

kschoche (approved):
LGTM

Issue #295, opened 2021-09-07
Annotation for error_on_missing_key


iusergii:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to set Vault Agent Template Config for [error_on_missing_key](https://www.vaultproject.io/docs/agent/template-config#interaction-between-exit_on_retry_failure-and-error_on_missing_key) parameters thru annotation. 

**Describe the solution you'd like**
Annotatio similar to [exit_on_retry_failure](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-template-config-exit-on-retry-failure)

tvoran:
Thanks @iusergii, this does seem like a good thing to add. Since it's a per-template setting, we'd probably want it to be implemented similar to [agent-inject-secret](https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-inject-secret) and [agent-inject-template](https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-inject-template), perhaps something like `error-on-missing-<name>`.

justinas-b:
+1

tvoran:
Added in #441

PR #441, opened 2023-03-03
feat annotations: error-on-missing-key annotation

dengaleev:
Relates: https://github.com/hashicorp/vault-k8s/issues/295

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -316,6 +316,13 @@ const (
 	// The name of the parameter is any unique string after "vault.hashicorp.com/agent-telemetry-",
 	// such as "vault.hashicorp.com/agent-telemetry-foobar".
 	AnnotationAgentTelemetryConfig = "vault.hashicorp.com/agent-telemetry"
+
+	// AnnotationErrorOnMissing is the key of annotation that configures whether
+	// template should error when a key is missing in the secret. The name of the
+	// secret is the string after "vault.hashicorp.com/error-on-missing-key-", and
+	// should map to the same unique value provided in
+	// "vault.hashicorp.com/agent-inject-secret-". Defaults to false
+	AnnotationErrorOnMissingKey = "vault.hashicorp.com/error-on-missing-key"
 )
 
 type AgentConfig struct {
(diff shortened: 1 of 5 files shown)


dengaleev:
Hi, @tvoran.
Could you look on this pull request?

matanz121:
Will appreciate merging this PR. We also need it as part of our vault usage.

shubhamsre:
@tvoran @dengaleev can help with merging this?

agent-inject/agent/annotations.go
```diff
@@ -642,6 +649,12 @@ func (a *Agent) secrets() []*Secret {
 				s.FilePermission = val
 			}
 
+			errMissingKey := fmt.Sprintf("%s-%s", AnnotationErrorOnMissingKey, raw)
+			if val, ok := a.Annotations[errMissingKey]; ok {
+				parsed, _ := strconv.ParseBool(val)
```
tvoran, line 654: We'll probably want to use `parseutil.ParseBool()` here to be consistent.

agent-inject/agent/annotations_test.go
```diff
@@ -628,6 +628,64 @@ func TestSecretCommandAnnotations(t *testing.T) {
...
+			map[string]string{
+				"vault.hashicorp.com/agent-inject-secret-foobar":  "test3",
+				"vault.hashicorp.com/error-on-missing-key-foobar": "false",
+			}, "foobar", false,
+		},
+		{
+			map[string]string{
+				"vault.hashicorp.com/agent-inject-secret-foobar": "test4",
+			}, "foobar", false,
+		},
```
tvoran, line 659: It would be good to have a test case where `error-on-missing-key-` is set to something besides true or false.

tvoran (commented):
Thanks! This is looking pretty good, and tested well for me locally.

It does need the latest changes from the `main` branch merged in, since much has changed in `Agent.secrets()` especially. Let us know if you'd like to continue working on this, otherwise we can finish it up.

agent-inject/agent/annotations_test.go
dengaleev: I've added a test to check a bad annotation value. But it only checks that an error occurs because `parseutil.ParseBool` doesn't wrap an error from `strconv`: https://github.com/mitchellh/mapstructure/blob/main/mapstructure.go#L741.

tvoran (approved):
Thanks!

PR #445, opened 2023-03-17
VAULT-14149 Don't override shareProcessNamespace unless annotation is explicitly set

### LINKED ISSUE #436: shareProcessNamespace is mutated even if "vault.hashicorp.com/agent-share-process-namespace" annotation is absent
**Describe the bug**

As part of https://github.com/hashicorp/vault-k8s/pull/334 and https://github.com/hashicorp/vault-k8s/pull/408 [vault.hashicorp.com/agent-share-process-namespace](http://vault.hashicorp.com/agent-share-process-namespace) annotation was implemented which mutates pod to have `shareProcessNamespace` property defined.

However it seems that if this annotation is not provided, it always defaults to `shareProcessNamespace: false` even if `shareProcessNamespace: true` is explicitly set on manifest.

**To Reproduce/Application deployment**

```
apiVersion: v1
kind: Namespace
metadata:
  name: reloading
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: reloading
  namespace: reloading
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: reloading
  namespace: reloading
  labels:
    app: reloading
spec:
  replicas: 1
  selector:
    matchLabels:
      app: reloading
  template:
    metadata:
      labels:
        app: reloading
      annotations:
        vault.hashicorp.com/agent-init-first: "true"
        vault.hashicorp.com/agent-run-as-same-user: "true"
        vault.hashicorp.com/template-static-secret-render-interval: 10s
        vault.hashicorp.com/agent-inject-command-MYSECRET: sh -c 'kill -HUP $(ps | grep -v grep | grep secret | grep -Eow "\d*" | head -1)'
        vault.hashicorp.com/agent-inject-file-MYSECRET: secret.txt
        vault.hashicorp.com/secret-volume-path-MYSECRET: /app
        vault.hashicorp.com/agent-inject-secret-MYSECRET: kv/data/reloading/reloading
        vault.hashicorp.com/agent-inject-template-MYSECRET: |
          {{- with secret "kv/data/reloading/reloading" -}}
          {{ .Data.data.foo }}
          {{- end }}
        vault.hashicorp.com/namespace: dev/vault-ne
        vault.hashicorp.com/role: service
    spec:
      serviceAccount: reloading
      serviceAccountName: reloading
      shareProcessNamespace: true
      containers:
      - image: ubuntu
        imagePullPolicy: IfNotPresent
        name: echoserver
        command: [ '/bin/sh', '-c', '--' ]
        args: [ 'content=$(cat /app/secret.txt) && while true; do echo $content; sleep 30; done;' ]
        ports:
        - containerPort: 80
        securityContext:
          runAsUser: 1000
```
When deploying this, we end up having `shareProcessNamespace` set to `false` after mutation, even if it is explicitly set to `true` in manifest

**Expected behavior**

I would expect `shareProcessNamespace` to property to be mutated only if pod has [vault.hashicorp.com/agent-share-process-namespace](http://vault.hashicorp.com/agent-share-process-namespace) annotation

**Environment**

* Kubernetes version:  EKS, GKE, AKS
* vault-k8s version: `hashicorp/vault-k8s:1.2.0`
* side-car:  `hashicorp/vault:1.9.2`

swenson:
By having a default of `false`, we would automatically set `shareProcessNamespace` to `false` when injecting the agent pod, regardless of what the main manifest had set.

Instead, we should only set `shareProcessNamespace` if the annotation is explicitly set, and otherwise let the default from the main manifest stand.

Fixes #436 (thanks @justinas-b for reporting it).

I tested this by:

* Before this patch:
  * Create a pod with `shareProcessNamespace: true` with injected agent
  * Verifying that the containers cannot see each other with `ps -ax`
* After this patch:
  * Create a pod with `shareProcessNamespace: true` with injected agent and no annotation
  * Verifying that the containers *can* see each other with `ps -ax`
  * Create a pod with `shareProcessNamespace: true` with injected agent and annotation set to `'false'`
  * Verifying that the containers *cannot* see each other with `ps -ax`


benashz (approved):
LGTM!

agent-inject/handler_test.go
tvoran, line 1: Might be nice to add some tests in here for shareProcessNamespace?

tvoran (approved):

swenson:
Thanks!

Issue #83, opened 2020-02-20
custom CA for external vault server


dontreboot:
AGENT_INJECT_VAULT_ADDR allows us to specify an external vault server address; however, I'd like also to add a custom CA to validate the vault server TLS. How do I specify that for vault injector?

jkaosw:
Discovered this today. This can be done via a Secret in Kubernetes and some annotations.  In the namespace you're spawning applications consuming the injector, there should be a Secret with the CA cert in it:

`kubectl create secret generic tls-ca --from-file=ca.pem`

And then for the annotation:

```
...
vault.hashicorp.com/tls-secret: "tls-ca"
vault.hasicorp.com/ca-cert: "/vault/tls/ca.pem"
...
```
How it works is that the annotation `vault.hashicorp.com/tls-secret` will mount the Kubernetes secret (e.g. tls-ca) to `/vault/tls` and then you reference in the `vault.hashicorp/ca-cert annotation.`

This is hidden away in the documentation under `vault.hasicorp.com/tls-secret.` Hope that's what you're looking for.

sbibow:
For me @jkaosw's solution only worked when I provided the complete certificate chain for my vault server. Using just the CA did not work.

So the secret was created with a file with this format:
```text
-----BEGIN CERTIFICATE-----
Vault certificate
...
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
Intermediate certificate
...
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
Root certificate
...
-----END CERTIFICATE-----
```

tomhjp:
Closed in #507, which adds the complementary `AGENT_INJECT_VAULT_CACERT_BYTES`.

Issue #19, opened 2019-12-21
Mechanism for injecting vault-ca.crt


cpick:
Very excited for this `vault`/k8s integration!

In our setup we have `vault` installed within its own, "vault", namespace.  It has a cluster-readable ConfigMap ("cm/vault-ca") that contains ".data.ca\.crt".  Currently, any Deployements/Jobs that would like to fetch a secret from `vault` GET "cm/vault-ca" from the k8s API Server and then use that to to TLS validation when communicating with `vault`.

With the injector I see that there is a `vault.hashicorp.com/tls-secret` annotation that will mount a Secret containing TLS certs (and that mount point can be referenced by `vault.hashicorp.com/ca-cert`), but since our "vault-ca" ConfigMap isn't a Secret (and, more importantly, is in a different namespace) we can't successfully mount it.  We could try to duplicate "cm/vault-ca"s contents into Secrets in each Namespace, but then we'd have to worry about keeping them in sync and about how to protect them from tampering.

Since we've already installed the injector alongside `vault` in the "vault" namespace and since it's already modifying Pods as they're created, it seems like it would be able to actually insert the correct CA into the `vault-agent` init containers/sidecars (much like it already does with the $VAULT_CONFIG environment variable).

jasonodonnell:
Hi @cpick, I think it could be a nice feature if we allowed setting the CA as an environment variable as part of the injection instead of mounting a secret.  We'll consider doing this in the new year!  For now you'll probably need to duplicate the CA until we support this.

eam-cs:
Any updates on this?

joemiller:
Interested in this feature as well. Might also be able to send a PR for it, time permitting.

@jasonodonnell Do you envision setting the contents of a PEM-encoded CA cert in the agent-injector's deployment.yaml and then injecting that as a file in the vault agent initContainer + sidecar?

perhaps something along the lines of:

```yaml
# agent-injector-deployment.yaml:
...
          args:
            - agent-inject
            - 2>&1
          env:
            - name: AGENT_INJECT_VAULT_CACERT
              value: |
                 -----BEGIN CERTIFICATE-----
                 ...
                 -----END CERTIFICATE-----
```

or perhaps it would be consumed via a file, possibly a ConfigMap or SecretVolume mounted into the agent injector container:

```yaml
# agent-injector-deployment.yaml:
...
          args:
            - agent-inject
            - 2>&1
          env:
[...]

PR #446, opened 2023-03-18
Adding support for the webhook injector to be able to inject a cert into the mutated Pods using an environment var.

gdasson:
* This PR adds the option for the webhook injector to be able to inject a cert into the mutated Pods using an environment var.
* This PR is an implementation of what has been suggested and explained very well [here](https://github.com/hashicorp/vault-k8s/issues/19#issuecomment-658436641) by @joemiller 
* The newly introduced env variable for injector: AGENT_INJECT_VAULT_CACERT_VALUE  could be supplied with a base64 encoded certificate string or plaintext. 
* This PR provides a solution for the below issues/questions:
https://github.com/hashicorp/vault-helm/issues/339
https://github.com/hashicorp/vault-k8s/issues/19
https://github.com/hashicorp/vault-helm/issues/777
https://github.com/hashicorp/vault-k8s/issues/83


aboukhal:
Until this very desirable PR gets merged, this is a valid work-around.

Dockerfile:
```
FROM hashicorp/vault:1.12.1
ENV VAULT_CAPATH /vault/tls/ca-cert.pem
COPY my-vault-certificate-chain.pem /vault/tls/ca-cert.pem
```

And then using that image in the helm chart values at `injector`.`agentImage`.

alnet:
I would also like to see this merged as we've just run into this exact issue today when building out a k3s cluster that will use an externally hosted vault. We'll be looking into the work around, but it's a shame to have to maintain a local image for such a tiny change.

tomhjp:
Thanks again for the PR - #507 is merged now and should be released pretty soon.

PR #487, opened 2023-06-14
update go, actions, k8s and vault versions

tvoran:
Updates all github actions to latest trusted versions and use the common jira-sync. Testing with k8s 1.23-1.27 (now that 1.22 has been dropped from major cloud providers). Building with go 1.20.5. Defaults to Vault 1.13.3.


tvoran:
Renamed the branch and re-opened in #488

PR #494, opened 2023-07-05
Add local testing Makefile targets

tomhjp:
Just some scripting to help smooth the process of testing local changes in a kind cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,15 +31,34 @@ build:
 image: build
 	docker build --build-arg VERSION=$(VERSION) --no-cache -t $(IMAGE_TAG) .
 
-deploy:
+deploy: image
 	kind load docker-image hashicorp/vault-k8s:$(VERSION)
 	helm upgrade --install vault vault --repo https://helm.releases.hashicorp.com --version=$(HELM_CHART_VERSION) \
 		--wait --timeout=5m \
 		--set 'server.dev.enabled=true' \
+		--set 'server.logLevel=debug' \
 		--set 'injector.image.tag=$(VERSION)' \
 		--set 'injector.image.pullPolicy=Never' \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,22 +14,64 @@ PKG=github.com/hashicorp/vault-k8s/version
...
-		-a \
 		-ldflags $(LDFLAGS) \
 		-o $(BUILD_DIR)/$(BIN_NAME) \
 		.
 
 image: build
 	docker build --build-arg VERSION=$(VERSION) --no-cache -t $(IMAGE_TAG) .
 
+# Deploys Vault dev server and a locally built Agent Injector.
+# Run multiple times to deploy new versions of the injector.
```
tvoran, line 35: Just a thought on wording:

```suggestion
# Run multiple times to deploy new builds of the injector.
```

Makefile
```diff
@@ -14,22 +14,64 @@ PKG=github.com/hashicorp/vault-k8s/version
...
+
+.PHONY: all test build image clean version deploy exercise teardown
 all: build
 
 version:
 	@echo $(VERSION)
 
 build:
 	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build \
-		-a \
```
tvoran, line 25: I'm curious about what motivated this change.

tvoran (approved):

Makefile
tomhjp: With all the k8s dependencies, the injector takes a while to rebuild from scratch, but I'm confident in Go's build caching correctness so I wanted to use the cache for speed. If we have any concerns about using the build cache though, we should probably disable the cache from the setup-go step in the CRT build so that it's still doing a build from scratch everytime.

PR #496, opened 2023-07-09
Vault sidecar type annotation

liad5h:


agent-inject/agent/container_sidecar_test.go
```diff
@@ -285,8 +285,9 @@ func TestContainerSidecar(t *testing.T) {
 		t.Errorf("wrong number of args, got %d, should have been %d", len(container.Args), 1)
 	}
 
-	if container.Args[0] != DefaultContainerArg {
-		t.Errorf("arg value wrong, should have been %s, got %s", DefaultContainerArg, container.Args[0])
+	arg := fmt.Sprintf(DefaultContainerArg, agent.SidecarType)
+	if container.Args[0] != arg {
+		t.Errorf("arg value wrong, should have been %s, got %s", arg, container.Args[0])
```
thyton, line 290: Could we also add a similar assertion for `agent.ProxyUseAutoAuthToken`?

agent-inject/agent/container_sidecar_test.go
```diff
@@ -285,8 +285,9 @@ func TestContainerSidecar(t *testing.T) {
 		t.Errorf("wrong number of args, got %d, should have been %d", len(container.Args), 1)
 	}
 
-	if container.Args[0] != DefaultContainerArg {
-		t.Errorf("arg value wrong, should have been %s, got %s", DefaultContainerArg, container.Args[0])
+	arg := fmt.Sprintf(DefaultContainerArg, agent.SidecarType)
```
thyton, line 288: Could you add different test cases under `TestContainerSidecar()` to test varying annotation sets containing 2 new annotations?
https://github.com/hashicorp/vault-k8s/blob/412e109929d5381bd77951956e7665fe8baaeaac/agent-inject/agent/annotations.go#L20-L28
https://github.com/hashicorp/vault-k8s/blob/412e109929d5381bd77951956e7665fe8baaeaac/agent-inject/agent/container_sidecar_test.go#L210-L216

thyton (changes requested):
@liad5h Thank you for taking this on and being patient! The PR is generally on a good path :). I have some change requests. I'm happy to sync up and help out with addressing them if you need.
 
Would you mind dropping a small description in the PR and something like below in CHANGELOG.md under Unreleased?
```
Improvements:
* Add support to configure sidecar type (proxy or agent) and proxy's auto auth token use through Pod Annotations [GH-496](https://github.com/hashicorp/vault-k8s/pull/496)
```

rathberm:
@thyton is this the official pr for this topic? Vault agent api proxy will be removed in 2026 Q2, I think this will break a lot of use-cases of this project. 
I'd be willing to take this on if you are not working on it already.

PR #505, opened 2023-07-30
Possibility to use template and template-file annotations without sec…

saturn4er:
## Summary
This Pull Request introduces enhancements to the Vault Kubernetes integration by simplifying the secret injection process. Specifically, it removes the need for the redundant and ineffective `vault.hashicorp.com/agent-inject-secret-foo: someKey` annotation.

## Motivation
In the current implementation, users are required to include the `vault.hashicorp.com/agent-inject-secret-foo: someKey` annotation in their configuration. However, this annotation doesn't contribute to the functionality and only adds unnecessary complexity.

## Changes Introduced
With this PR, users will no longer need to specify the `vault.hashicorp.com/agent-inject-secret-foo: someKey` annotation. Instead, the necessary information about secrets can be fully determined by the provided templates or template files, such as `vault.hashicorp.com/agent-inject-template-foo: "{{- with secret "someKey" -}} {{- end }}"` or `vault.hashicorp.com/agent-inject-template-file-foo: "someFile"`. 
This results in a cleaner and more intuitive configuration.

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -589,58 +589,77 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 func (a *Agent) secrets() []*Secret {
 	var secrets []*Secret
 
-	for name, path := range a.Annotations {
-		secretName := fmt.Sprintf("%s-", AnnotationAgentInjectSecret)
-		if strings.Contains(name, secretName) {
-			raw := strings.ReplaceAll(name, secretName, "")
-			name := raw
-
-			if ok, _ := a.preserveSecretCase(raw); !ok {
-				name = strings.ToLower(raw)
-			}
-
-			if name == "" {
+	for annotationName, annotationValue := range a.Annotations {
+		var (
+			secret    *Secret
+			raw, name string
+			ok        bool
+		)
+		switch {
+		// parsing vault.hashicorp.com/agent-inject-secret-<name>: <path>
+		case strings.Contains(annotationName, fmt.Sprintf("%s-", AnnotationAgentInjectSecret)):
+			raw, name, ok = a.secretName(annotationName, AnnotationAgentInjectSecret)
+			if !ok {
 				continue
 			}
-
-			s := &Secret{Name: name, Path: path}
-
-			templateName := fmt.Sprintf("%s-%s", AnnotationAgentInjectTemplate, raw)
-			if val, ok := a.Annotations[templateName]; ok {
-				s.Template = val
+			secret = &Secret{Name: name, Path: annotationValue}
+		// parsing vault.hashicorp.com/agent-inject-secret-template-file-<name>: <template-file>
+		case strings.Contains(annotationName, fmt.Sprintf("%s-", AnnotationAgentInjectTemplateFile)):
+			raw, name, ok = a.secretName(annotationName, AnnotationAgentInjectTemplateFile)
+			if !ok {
+				continue
 			}
-			if s.Template == "" {
-				templateFileAnnotation := fmt.Sprintf("%s-%s", AnnotationAgentInjectTemplateFile, raw)
-				if val, ok := a.Annotations[templateFileAnnotation]; ok {
-					s.TemplateFile = val
-				}
+			secret = &Secret{Name: name, TemplateFile: annotationValue}
+		// parsing vault.hashicorp.com/agent-inject-secret-template-<name>: <template>
+		case strings.Contains(annotationName, fmt.Sprintf("%s-", AnnotationAgentInjectTemplate)):
+			raw, name, ok = a.secretName(annotationName, AnnotationAgentInjectTemplate)
+			if !ok {
+				continue
 			}
-
-			s.MountPath = a.Annotations[AnnotationVaultSecretVolumePath]
-			mountPathAnnotationName := fmt.Sprintf("%s-%s", AnnotationVaultSecretVolumePath, raw)
-			if val, ok := a.Annotations[mountPathAnnotationName]; ok {
-				s.MountPath = val
-			}
-
-			commandName := fmt.Sprintf("%s-%s", AnnotationAgentInjectCommand, raw)
-			if val, ok := a.Annotations[commandName]; ok {
-				s.Command = val
-			}
-
-			file := fmt.Sprintf("%s-%s", AnnotationAgentInjectFile, raw)
-			if val, ok := a.Annotations[file]; ok {
-				s.FilePathAndName = val
-			}
-
-			filePerm := fmt.Sprintf("%s-%s", AnnotationAgentInjectFilePermission, raw)
-			if val, ok := a.Annotations[filePerm]; ok {
-				s.FilePermission = val
-			}
-
-			secrets = append(secrets, s)
+			secret = &Secret{Name: name, Template: annotationValue}
+		default:
+			continue
 		}
+
+		secret.MountPath = a.Annotations[AnnotationVaultSecretVolumePath]
+		mountPathAnnotationName := fmt.Sprintf("%s-%s", AnnotationVaultSecretVolumePath, raw)
+		if val, ok := a.Annotations[mountPathAnnotationName]; ok {
+			secret.MountPath = val
+		}
+
+		commandName := fmt.Sprintf("%s-%s", AnnotationAgentInjectCommand, raw)
+		if val, ok := a.Annotations[commandName]; ok {
+			secret.Command = val
+		}
+
+		file := fmt.Sprintf("%s-%s", AnnotationAgentInjectFile, raw)
+		if val, ok := a.Annotations[file]; ok {
+			secret.FilePathAndName = val
+		}
+
+		filePerm := fmt.Sprintf("%s-%s", AnnotationAgentInjectFilePermission, raw)
+		if val, ok := a.Annotations[filePerm]; ok {
+			secret.FilePermission = val
+		}
+
+		secrets = append(secrets, secret)
 	}
 	return secrets
+}
+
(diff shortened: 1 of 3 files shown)


agent-inject/agent/annotations.go
```diff
@@ -589,60 +589,79 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
-		if strings.Contains(name, secretName) {
-			raw := strings.ReplaceAll(name, secretName, "")
-			name := raw
-
-			if ok, _ := a.preserveSecretCase(raw); !ok {
-				name = strings.ToLower(raw)
-			}
-
-			if name == "" {
+	for annotationName, annotationValue := range a.Annotations {
```
tomhjp, line 592: I think if someone specifies multiple annotations from the `switch` e.g. _both_ `vault.hashicorp.com/agent-inject-secret-template-file-<name>` and `vault.hashicorp.com/agent-inject-secret-<name>`, we'll add two entries to the `secrets` slice. I'm not sure how that will shake out further along in the execution path, but it would be good to add a test and guard against that to avoid unexpected behaviour.

agent-inject/agent/annotations_test.go
```diff
@@ -348,44 +349,30 @@ func TestSecretLocationFileAnnotations(t *testing.T) {
...
 		{
 			map[string]string{
-				"vault.hashicorp.com/agent-inject-secret-foobar":  "test1",
 				"vault.hashicorp.com/agent-inject-templat-foobar": "foobarTemplate",
-			}, "foobar", "",
-		},
-		{
-			map[string]string{
-				"vault.hashicorp.com/agent-inject-secret-foobar":    "test1",
-				"vault.hashicorp.com/agent-inject-template-foobar2": "foobarTemplate",
```
tomhjp, line 375: Ah I see you deleted these test cases, which relates to my comment above. To preserve backwards compatibility, I think it's still important that we handle these cases the same as before, which shouldn't require too much effort in the implementation.

agent-inject/agent/annotations_test.go
```diff
@@ -424,24 +415,20 @@ func TestSecretMixedTemplatesAnnotations(t *testing.T) {
 	}{
 		{
 			map[string]string{
-				"vault.hashicorp.com/agent-inject-secret-foobar":        "test1",
-				"vault.hashicorp.com/agent-inject-template-foobar":      "",
```
tomhjp, line 428: Please can we leave these test cases (here and below) in place and add new ones for the new behaviour?

agent-inject/agent/config_test.go
```diff
@@ -29,7 +29,6 @@ func TestNewConfig(t *testing.T) {
 		AnnotationVaultClientKey:                        "client-key",
 		AnnotationVaultSecretVolumePath:                 "/vault/secrets",
 		AnnotationProxyAddress:                          "http://proxy:3128",
-		"vault.hashicorp.com/agent-inject-secret-foo":   "db/creds/foo",
```
tomhjp, line 32: Same comment again - it would be good to preserve existing behaviour and just enable a new simplified set of annotations that previously wouldn't have worked.

tomhjp (commented):
I really like the concept for this, and would love to get it merged. My comments just resolve around wanting to preserve backwards compatibility so we don't break existing users. We can discuss a breaking change if you think it's necessary, but my read is that it shouldn't be too hard to avoid it.

agent-inject/agent/annotations.go
saturn4er: So I thought about that, and the only one solution I found is to prioritize file template and template annotation over secret annotation, so injector will skip `template` if `file-template` specified, and also skip `secret` if `template` specified

saturn4er: @tomhjp what do you think?

saturn4er:
@tomhjp Ready, old test cases are now works and I added new ones to cover case when only template or template-file annotation set up

agent-inject/agent/annotations.go
```diff
@@ -587,62 +587,100 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
-			s.MountPath = a.Annotations[AnnotationVaultSecretVolumePath]
-			mountPathAnnotationName := fmt.Sprintf("%s-%s", AnnotationVaultSecretVolumePath, raw)
-			if val, ok := a.Annotations[mountPathAnnotationName]; ok {
-				s.MountPath = val
-			}
+	for _, secret := range templateFileSecrets {
+		if templateSecrets[secret.Name] == nil {
+			secrets = append(secrets, secret)
+		}
+	}
```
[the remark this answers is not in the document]
saturn4er: So if we're setting Path unconditionally, why we should set TemplateFile conditionally? I pushed some changes, please check if you're ok with them

saturn4er:
@tomhjp any issues with this implementation?

agent-inject/agent/annotations_test.go
```diff
@@ -495,7 +526,7 @@ func TestSecretTemplateFileAnnotations(t *testing.T) {
 				"vault.hashicorp.com/agent-inject-secret-foobar":        "test1",
 				"vault.hashicorp.com/agent-inject-template-foobar":      "foobarTemplate",
 				"vault.hashicorp.com/agent-inject-template-file-foobar": "/etc/config.tmpl",
-			}, "foobar", "foobarTemplate", "",
+			}, "foobar", "foobarTemplate", "/etc/config.tmpl",
```
tomhjp, line 529: I don't think this change makes sense in the context of what the test is trying to assert. And I think it's highlighting a backwards incompatibility. I'll leave another comment on the actual fix site.

agent-inject/agent/annotations.go
```diff
@@ -587,62 +587,73 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
-				s.FilePermission = val
-			}
+	}
 
-			secrets = append(secrets, s)
-		}
+	for _, secret := range secrets {
+		secret.Path = a.annotationsSecretValue(AnnotationAgentInjectSecret, secret.RawName, secret.Path)
+		secret.Template = a.annotationsSecretValue(AnnotationAgentInjectTemplate, secret.RawName, secret.Template)
+		secret.TemplateFile = a.annotationsSecretValue(AnnotationAgentInjectTemplateFile, secret.RawName, secret.TemplateFile)
```
tomhjp, line 626:
```suggestion
	if secret.Template == "" {
		secret.TemplateFile = a.annotationsSecretValue(AnnotationAgentInjectTemplateFile, secret.RawName, secret.TemplateFile)
	}
```

I know you disagreed with this elsewhere, but I still think this is worth doing. Here's where the downstream processing of `secret.Template` and `secret.TemplateFile` takes place:

https://github.com/hashicorp/vault-k8s/blob/02af6ab44ed3c54c3ecf503593901755b3d72799/agent-inject/agent/config.go#L180-L214

The logic is kind of confusingly written, but take the case where both the template and template-file annotations are non-empty. Previously, `secret.Template` is non-empty and `secret.TemplateFile` is empty, and the same happens in `tmpl`. Now, _both_ are non-empty in `secret` and the same in `tmpl`. Maybe we could chase the behaviour of what happens next all the way down into consul-template, but I'd rather just take the high confidence route of keeping the localised behaviour the same here.

agent-inject/agent/annotations.go
```diff
@@ -587,62 +587,73 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
-			filePerm := fmt.Sprintf("%s-%s", AnnotationAgentInjectFilePermission, raw)
-			if val, ok := a.Annotations[filePerm]; ok {
-				s.FilePermission = val
-			}
+	}
 
-			secrets = append(secrets, s)
-		}
+	for _, secret := range secrets {
+		secret.Path = a.annotationsSecretValue(AnnotationAgentInjectSecret, secret.RawName, secret.Path)
```
tomhjp, line 624: It reads a little strangely to me to use the values of fields we haven't set yet as the default. Could we just pass in "" as the default for all the lines except volume path to make it a little easier to read?

```suggestion
		secret.Path = a.annotationsSecretValue(AnnotationAgentInjectSecret, secret.RawName, "")
```

tomhjp (commented):
Sorry for the delay doing another round of reviewing. This looks really close now, one change that I think is a requirement and the other comments are just nits.

tomhjp:
@saturn4er I'm going to start kicking off a release soon, but I'd love to get this merged first. LMK if you're able to address the remaining comments - I pushed a commit that implements the comments here if that's helpful: 92153f636e4a43b82ca8d5f6458042457537e08b

agent-inject/agent/annotations.go
saturn4er: Yeah, I wrote like this just in case if there will be some default values setted at the top, so this one will not overwrite them.

agent-inject/agent/annotations_test.go
saturn4er: As you wish :)

agent-inject/agent/annotations_test.go
```diff
@@ -389,31 +392,37 @@ func TestSecretTemplateAnnotations(t *testing.T) {
...
-		}
+			agent, err := New(pod)
+			if err != nil {
+				t.Errorf("got error, shouldn't have: %s", err)
+			}
 
-		if len(agent.Secrets) == 0 {
-			t.Error("Secrets length was zero, it shouldn't have been")
-		}
+			var found bool
```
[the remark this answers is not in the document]
saturn4er: I rewrote it to maps

saturn4er:
@tomhjp, I've made revisions based on your suggestions. Could you please review the changes?

PR #532, opened 2023-10-12
go, k8s, and actions updates

tvoran:
Build with go 1.21.3 and update related go packages. Pin github actions to latest trusted versions, and test with k8s 1.24-1.28.


benashz (approved):
LGTM!

Issue #544, opened 2023-11-01
Injected config tries to use IRSA token instead of the k8s service account token.


cm3lindsay:
**Describe the bug**
When the Vault Agent sidecar is injected into a pod that is also injected with IRSA, the token_path uses the IRSA token instead of the actual Kubernetes service account token even when configured to use the Kubernetes auth method. This prevents the Vault Agent sidecar from being able to authenticate to Vault.

**To Reproduce**
1. Deploy pod annotated for vault-agent injection that is configured to use a service account that has an IRSA annotation.
2. Vault Agent logs will show failed authentication requests
3. Inspecting the containers VAULT_CONFIG env variable will show that `token_path` is configured to use `/var/run/secrets/eks.amazonaws.com/serviceaccount` instead of `var/run/secrets/kubernetes.io/serviceaccount`.

**Expected behavior**
Vault agent is able to authenticate to Vault using the Kubernetes auth method without having to specify `vault.hashicorp.com/auth-config-token-path` to ensure the wrong token is not used.

**Environment**
* Kubernetes version: 
  * EKS 1.24
* vault-k8s version:
  * 1.3.1

**Additional context**
We discovered this problem after configuring the Vault Injector to use an external address.

ArieLevs:
wasted couple of hours on debugging vault authentication issues, until i found this issue :rocket:

was migrating vault agent to a shot-lived token authentication from a static token auth, which worked totally OK,
having other applications using shot-lived token with kubernetes auth method, also worked totally OK so was not able to understand what is wrong with vault agent.

removing IRSA annotation from my test application totally made everything work like it should again, obviously removing this annotation is no a solution.

is there an estimation of when above PR gets merged?

PR #545, opened 2023-11-01
fix: prevent using IRSA token for Kubernetes auth

cm3lindsay:
Summary
Skip IRSA token when detecting the service account token to use for the Kubernetes auth method.

I tried to follow the same logic used elsewhere for detecting the IRSA token.

Fixes
https://github.com/hashicorp/vault-k8s/issues/544


RyanW8:
We've encountered a similar issue but instead with EKS Pod Identities and raised https://github.com/hashicorp/vault-k8s/pull/793/files

PR #579, opened 2024-01-18
Ccapurso/injector max conns per host

ccapurso:


agent-inject/agent/annotations.go
```diff
@@ -519,6 +525,14 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval] = cfg.StaticSecretRenderInterval
 	}
 
+	//if maxConnsPerHostRaw, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost]; !ok {
```
VioletHynes, line 528: Is this meant to be commented out?

agent-inject/agent/annotations.go
```diff
@@ -812,6 +826,15 @@ func (a *Agent) templateConfigExitOnRetryFailure() (bool, error) {
 	return strconv.ParseBool(raw)
 }
 
+func (a *Agent) templateConfigMaxConnectionsPerHost() (int64, error) {
+	raw, ok := a.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost]
+	if !ok {
+		return DefaultTemplateConfigMaxConnectionsPerHost, nil
+	}
+
+	return strconv.ParseInt(raw, 10, 64)
```
VioletHynes, line 835: Would `parseutil.ParseInt` be better here?

VioletHynes: I know we try and use it where we can, but I don't mind if we think it's overkill

agent-inject/agent/annotations.go
ccapurso: Nope, I forgot to add my actual changes. Whoops! Thank you.

agent-inject/agent/annotations.go
ccapurso: I was on the fence about that and chose to remain consistent with the current implementation. I like the improvement, however, and will make the change. Thanks!

agent-inject/agent/agent.go
```diff
@@ -7,36 +7,37 @@ import (
...
+	DefaultAgentSetSecurityContext             = true
+	DefaultAgentReadOnlyRoot                   = true
+	DefaultAgentCacheEnable                    = "false"
+	DefaultAgentCacheUseAutoAuthToken          = "true"
+	DefaultAgentCacheListenerPort              = "8200"
+	DefaultAgentCacheExitOnErr                 = false
+	DefaultAgentUseLeaderElector               = false
+	DefaultAgentInjectToken                    = false
+	DefaultTemplateConfigExitOnRetryFailure    = true
+	DefaultTemplateConfigMaxConnectionsPerHost = 10
```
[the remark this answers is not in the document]
ccapurso: Agent does have the same default. I erred on the side of redundancy. I'm happy to remove this if that is preferred.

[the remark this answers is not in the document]
tvoran: Up to you, just seems unnecessary to set a default on the injector side as well. Though I do see that setting `AGENT_INJECT_TEMPLATE_MAX_CONNECTIONS_PER_HOST="0"` effectively removes the default if it becomes an issue.

agent-inject/agent/annotations.go
```diff
@@ -519,6 +526,14 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigStaticSecretRenderInterval] = cfg.StaticSecretRenderInterval
 	}
 
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost]; !ok {
+		if cfg.MaxConnectionsPerHost == 0 {
+			cfg.MaxConnectionsPerHost = DefaultTemplateConfigMaxConnectionsPerHost
+		}
+
+		pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost] = strconv.FormatInt(cfg.MaxConnectionsPerHost, 10)
```
tvoran, line 534: I'm not that familiar with the parseutil package, so I'm curious why it's being used some places and not others (like this line).

ccapurso: I believe that we're trying to make an effort on using `parseutil` within `go-secure-stdlib` for arbitrarily-typed values. For example, [ParseInt](https://github.com/hashicorp/go-secure-stdlib/blob/b4a7f3e5652cb93b2c3c3f42bb2b2c8ed21641ac/parseutil/parseutil.go#L215) handles strings, `json.Number`, and various integer types. I do not believe that `go-secure-stdlib` has an analogous function to format integers as strings. In this case, we're providing a hardcoded scalar value so it is more deterministic.

agent-inject/agent/agent.go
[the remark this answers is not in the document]
ccapurso: Sure, I can remove. Having both creates a maintenance burden as well which is not ideal. Thanks for the feedback!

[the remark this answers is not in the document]
ccapurso: The default has been removed in commit https://github.com/hashicorp/vault-k8s/pull/579/commits/a777fcae2d982d8f8bbc67f0dcb5439a8a6daa36.

agent-inject/agent/annotations.go
tvoran: It looks like `parseutil.ParseString()` may work here? But good point about it not being arbitrary input at this point, that makes sense.

ccapurso: Oh that's a good point. It makes sense to remain consistent since that was the whole point of the change.

ccapurso: Actually, on second thought, modifying it requires additional weak decode and error handling so I will leave it as is.

agent-inject/agent/agent.go
```diff
@@ -7,36 +7,36 @@ import (
...
-	DefaultAgentCacheEnable                 = "false"
-	DefaultAgentCacheUseAutoAuthToken       = "true"
-	DefaultAgentCacheListenerPort           = "8200"
-	DefaultAgentCacheExitOnErr              = false
-	DefaultAgentUseLeaderElector            = false
-	DefaultAgentInjectToken                 = false
-	DefaultTemplateConfigExitOnRetryFailure = true
-	DefaultServiceAccountMount              = "/var/run/secrets/vault.hashicorp.com/serviceaccount"
-	DefaultEnableQuit                       = false
-	DefaultAutoAuthEnableOnExit             = false
```
tvoran, line 39: This indentation change can probably be reverted?

tvoran (approved):
Tested great for me locally too 👍 

Feel free to add a changelog entry under Unreleased -> Features too if you like.

PR #619, opened 2024-04-02
fix: disable handling update on pods

aslafy-z:


deploy/injector-mutating-webhook.yaml
```diff
@@ -22,10 +22,14 @@ webhooks:
...
     rules:
+      - operations: ["CREATE"]
+        apiGroups: [""]
+        apiVersions: ["v1"]
+        resources: ["pods"]
       - operations: ["CREATE", "UPDATE"]
         apiGroups: [""]
         apiVersions: ["v1"]
-        resources: ["deployments", "jobs", "pods", "statefulsets"]
+        resources: ["deployments", "jobs", "statefulsets"]
```
tomhjp, line 32: We can actually delete this whole entry as well - I'm not sure why we ever added this but it looks like a mistake to me. The first thing the handler does is try to unmarshal the object its given as a pod.

tomhjp (approved):
LGTM, thanks for raising this. IIUC this should mostly be fixed with https://github.com/hashicorp/vault-k8s/pull/501 which got released in v1.3.0, but it's nice to do this bit of cleanup as well.

I'll leave this over the weekend before merging in case you want to address the suggestion, but I'm also fine with merging as-is and addressing the comment in a follow-up.

deploy/injector-mutating-webhook.yaml
```diff
@@ -22,10 +22,10 @@ webhooks:
         namespace: "vault"
       caBundle: ""
     rules:
-      - operations: ["CREATE", "UPDATE"]
+      - operations: ["CREATE"]
         apiGroups: [""]
         apiVersions: ["v1"]
-        resources: ["deployments", "jobs", "pods", "statefulsets"]
+        resources: ["pods"]
```
tvoran, line 28: Would you mind adding `scope` here too? This was [recently added](https://github.com/hashicorp/vault-helm/pull/1037) to our helm chart and it would good to be consistent.

```suggestion
        resources: ["pods"]
        scope: "Namespaced"
```

tvoran (commented):
Just one more addition and I think we're good!

deploy/injector-mutating-webhook.yaml
aslafy-z: Applied!

tvoran (approved):

PR #623, opened 2024-04-04
Prepare for v1.4.1 release

tvoran:
Updating go, changelog, and default vault version.

Also uses bats v1.11.0.


thyton (approved):
LGTM!

tvoran:
Thanks!

.github/workflows/tests.yaml
```diff
@@ -39,10 +39,10 @@ jobs:
     steps:
       # Setup test tools from https://github.com/hashicorp/vault-helm/blob/main/.github/workflows/setup-test-tools/action.yaml
       - uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
-      - run: npm install -g bats@${BATS_VERSION}
+      - run: sudo npm install -g bats@${BATS_VERSION}
```
benashz, line 42: @tvoran Is the sudo required here because of the bats version bump?

tvoran: I think it's more something to do with a change in the runners. The bats version bump was passing, at least intermittently, but then started failing consistently due to permissions on /usr/local/share/man/ IIRC.

PR #661, opened 2024-07-01
Prepare release v1.4.2

tvoran:
Updating go, changelog, default vault version, and CI versions of vault and vault-helm. Also updates dependabot to group gomod updates and only update hashicorp github-actions.


benashz (approved):

tvoran:
Thanks!

PR #664, opened 2024-07-08
Delimiter annotation

Diliz:
Rebased @cailtlinelfring annotations branch with delimiter annotations feature


agent-inject/agent/annotations.go
```diff
@@ -82,6 +82,24 @@ const (
 
 	// AnnotationAgentInjectToken is the annotation key for injecting the
 	// auto-auth token into the secrets volume (e.g. /vault/secrets/token)
+	// AnnotationAgentInjectTemplateLeftDelim is the key annotation that configures Vault
+	// Agent what left delimiter to use for rendering the secrets.  The name
+	// of the template is any unique string after "vault.hashicorp.com/agent-inject-left-delimiter-",
+	// such as "vault.hashicorp.com/agent-inject-left-delimiter-foobar".  This should map
+	// to the same unique value provided in "vault.hashicorp.com/agent-inject-secret-".
+	// If not provided, a default left delimiter is used as defined by https://www.vaultproject.io/docs/agent/template#left_delimiter
+	AnnotationAgentInjectTemplateLeftDelim = "vault.hashicorp.com/agent-inject-left-delimiter"
```
Diliz, line 91: changed the inject-template-left-delimiter from inject-left-delimiter since the default pull request

Diliz:
fix https://github.com/hashicorp/vault-helm/issues/348 as well

Diliz:
@tvoran @benashz How much time does it take for pull requests from outsiders to be checked and merged? (Just to know if this will be available soon or not on my side)

benashz:
The original pull request talks a bit about handling templates included in a Helm chart. I wonder if we could recommend escaping the template instead of adding support for specifying alternate template delimiters? We are suggesting that approach here: https://github.com/hashicorp/vault-secrets-operator/issues/619#issuecomment-1984053146

Diliz:
> The original pull request talks a bit about handling templates included in a Helm chart. I wonder if we could recommend escaping the template instead of adding support for specifying alternate template delimiters? We are suggesting that approach here: [hashicorp/vault-secrets-operator#619 (comment)](https://github.com/hashicorp/vault-secrets-operator/issues/619#issuecomment-1984053146)

Humm, this helm approache is a workaround, not something to fix the real issue, currently it's a hassle to use go templating inside vault templates in kubernetes, if you want to make templating over templating over termplating, it becomes impossible to use if you don't know beforehand how many times the template will be templated, so escaping it is not a solution in this case.

Why not simply adding this delim annotations to make it convenient and be able to differentiate the default go templating from the one for the vault agent?

agent-inject/agent/annotations.go
tvoran: I was also thinking we could shorten it a little more and use something like `agent-template-left-delim` here? I like having "template" in the annotation key since this is a template option.

agent-inject/agent/annotations.go
```diff
@@ -646,6 +664,8 @@ func (a *Agent) secrets() ([]*Secret, error) {
 		}
 		secret.MountPath = a.annotationsSecretValue(AnnotationVaultSecretVolumePath, secret.RawName, a.Annotations[AnnotationVaultSecretVolumePath])
 		secret.Command = a.annotationsSecretValue(AnnotationAgentInjectCommand, secret.RawName, "")
+		secret.LeftDelimiter = a.annotationsSecretValue(AnnotationAgentInjectTemplateLeftDelim, secret.RawName, "")
+		secret.RightDelimiter = a.annotationsSecretValue(AnnotationAgentInjectTemplateRightDelim, secret.RawName, "")
```
tvoran, line 668: Seems like we might as well set the default delimiter here?

```suggestion
		secret.LeftDelimiter = a.annotationsSecretValue(AnnotationAgentInjectTemplateLeftDelim, secret.RawName, DefaultLeftDelim)
		secret.RightDelimiter = a.annotationsSecretValue(AnnotationAgentInjectTemplateRightDelim, secret.RawName, DefaultRightDelim)
```

tvoran (commented):
Thanks for updating this! It tested good locally for me; just left a couple thoughts.

Something else I think may be useful here is a way to set the default template delimiters for all the templates in a Pod, instead of per-template. Perhaps the delimiter annotations by themselves (without the secret suffix) could be used for this? Just something I was thinking about while reviewing this, but we can do it in a follow-up PR too.

agent-inject/agent/annotations.go
Diliz: The thing is, we can then makes the default delimiters handled by the config, which is where the templating will be done, it was done this way in the previous pull request, I did it the way you say in the first place but I think it's better/less error prone to let the config handle that, no?

agent-inject/agent/annotations.go
Diliz: Yep, I'll change that 👍

markush81:
Just stumbled over this and it is exactly what i need as well. Doing the escaping kinda gets easily a hell if you want to have parts replaced by helm template and parts by vault.

Any ETA for this PR?

Diliz:
> Just stumbled over this and it is exactly what i need as well. Doing the escaping kinda gets easily a hell if you want to have parts replaced by helm template and parts by vault.
> 
> Any ETA for this PR?

Waiting for @tvoran and @benashz comments and approval, I'm still following this case as I need it implemented as well 👍

agent-inject/agent/annotations.go
tvoran: Not sure I follow about it being better/less error prone? Since the delimiter annotations are set directly in the corresponding Agent config options (i.e. nothing is changing them in between parsing the annotations and writing the Agent config), seems like they can be set as soon as they are known.

If the delimiter settings end up empty in the Agent config because an annotation sets them to `""`, that's actually ok too since they aren't required Agent config parameters.

(I don't think it matters too much either way, so I'll leave it up to you and not hold it up over this.)

agent-inject/agent/annotations.go
```diff
@@ -82,6 +82,26 @@ const (
...
 	// AnnotationAgentInjectToken is the annotation key for injecting the
 	// auto-auth token into the secrets volume (e.g. /vault/secrets/token)
+	// AnnotationAgentInjectTemplateLeftDelim is the key annotation that configures Vault
+	// Agent what left delimiter to use for rendering the secrets.  The name
+	// of the template is any unique string after "vault.hashicorp.com/agent-template-left-delim-",
+	// such as "vault.hashicorp.com/agent-template-left-delim-foobar".  This should map
+	// to the same unique value provided in "vault.hashicorp.com/agent-inject-secret-".
+	// If not provided, a default left delimiter is used as defined by https://www.vaultproject.io/docs/agent/template#left_delimiter
+	AnnotationAgentInjectTemplateLeftDelim = "vault.hashicorp.com/agent-template-left-delim"
+	DefaultLeftDelim                       = "{{"
```
tvoran, line 92: I'd rather we set these `DefaultLeftDelim` and `DefaultRightDelim` in config.go; they'll be easier to spot there.

Diliz:
Made the requested changes, it should be good now :)

+ Rebases as well

agent-inject/agent/annotations.go
```diff
@@ -80,6 +80,24 @@ const (
 	// If not provided, the template content key annotation is used.
 	AnnotationAgentInjectTemplateFile = "vault.hashicorp.com/agent-inject-template-file"
 
+	// AnnotationAgentInjectToken is the annotation key for injecting the
+	// auto-auth token into the secrets volume (e.g. /vault/secrets/token)
```
tvoran, line 84:
```suggestion
```

agent-inject/agent/config_test.go
```diff
@@ -134,6 +139,10 @@ func TestNewConfig(t *testing.T) {
 			if template.Contents != "template foo" {
 				t.Errorf("expected template contents to be %s, got %s", "template foo", template.Contents)
 			}
+
+			if template.LeftDelim != DefaultLeftDelim || template.RightDelim != DefaultRightDelim {
+				t.Errorf("expected default delimiters to be %s (left) and %s (right), got %s (left) and %s (right)", template.LeftDelim, template.RightDelim, DefaultLeftDelim, DefaultRightDelim)
```
tvoran, line 144: Looks like the "got" and "expected" values are flipped:

```suggestion
				t.Errorf("expected default delimiters to be %s (left) and %s (right), got %s (left) and %s (right)", DefaultLeftDelim, DefaultRightDelim, template.LeftDelim, template.RightDelim)
```

agent-inject/agent/config_test.go
```diff
@@ -170,6 +179,10 @@ func TestNewConfig(t *testing.T) {
 			if template.Destination != "/vault/secrets/just-template-file" {
 				t.Errorf("expected template destination to be %s, got %s", "/vault/secrets/just-template-file", template.Destination)
 			}
+		} else if strings.Contains(template.Destination, "baz") {
+			if template.LeftDelim != "[[" || template.RightDelim != "]]" {
+				t.Errorf("expected default delimiters to be %s (left) and %s (right), got %s (left) and %s (right)", template.LeftDelim, template.RightDelim, "[[", "]]")
```
tvoran, line 184:
```suggestion
				t.Errorf("expected default delimiters to be %s (left) and %s (right), got %s (left) and %s (right)", "[[", "]]", template.LeftDelim, template.RightDelim)
```

tvoran (commented):
Just a couple more minor things.

agent-inject/agent/config_test.go
```diff
@@ -121,8 +126,8 @@ func TestNewConfig(t *testing.T) {
 		t.Error("agent Cache should be disabled for init containers")
 	}
 
-	if len(config.Templates) != 6 {
-		t.Errorf("expected 4 template, got %d", len(config.Templates))
+	if len(config.Templates) != 7 {
+		t.Errorf("expected 7 template, got %d", len(config.Templates))
```
tvoran, line 130: Good catch!

```suggestion
		t.Errorf("expected 7 templates, got %d", len(config.Templates))
```

tvoran (approved):

tvoran:
Thanks so much! We'll get this into the next release.

PR #679, opened 2024-08-30
added envvar for tls

kondotak:
I needed to add a ca-cert that matched a given cert and key.  Here is the code i added that provides cli arg and envvar arg... first time using golang, be gentle!

### DIFF
--- a/subcommand/injector/command.go
+++ b/subcommand/injector/command.go
@@ -47,6 +47,7 @@ type Command struct {
 	flagListen                     string // Address of Vault Server
 	flagLogLevel                   string // Log verbosity
 	flagLogFormat                  string // Log format
+	flagCACertFile                 string // TLS CA Certificate to serve
 	flagCertFile                   string // TLS Certificate to serve
 	flagKeyFile                    string // TLS private key to serve
 	flagExitOnRetryFailure         bool   // Set template_config.exit_on_retry_failure on agent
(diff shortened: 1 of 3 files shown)


subcommand/injector/flags.go
```diff
@@ -56,6 +56,9 @@ type Specification struct {
 	// TLSAutoHosts is the AGENT_INJECT_TLS_AUTO_HOSTS environment variable.
 	TLSAutoHosts string `envconfig:"tls_auto_hosts"`
 
+	// TLSCACertFile is the AGENT_INJECT_TLS_CA_CERT_FILE environment variable.
+	TLSCACertFile string `envconfig:"tls_ca_cert_file"`
```
tvoran, line 60: We'll probably want to add a test case for AGENT_INJECT_TLS_CA_CERT_FILE in [TestCommandEnvs()](https://github.com/kondotak/vault-k8s/blob/a160ece8060796d04732cb640cc5d393b4f7c05f/subcommand/injector/flags_test.go#L110) as well.

subcommand/injector/command.go
```diff
@@ -179,6 +180,7 @@ func (c *Command) Run(args []string) int {
 	}
 	if c.flagCertFile != "" {
 		certSource = &cert.DiskSource{
+			CAPath:  c.flagCACertFile,
```
tvoran, line 183:
```suggestion
			CAPath:   c.flagCACertFile,
```

tvoran (commented):
Thanks for submitting this! I think if we update unit test I mentioned it should be good to go.

subcommand/injector/flags.go
kondotak: okay, added test case, easy to add!

tvoran (approved):
Thanks! We'll get this into the next release.

PR #698, opened 2024-11-02
Add annotation for an agent metrics listener

LS80:


LS80:
@tvoran Before I go any further with this, is it something that is likely to be accepted?

plieberg:
> @tvoran Before I go any further with this, is it something that is likely to be accepted?

If so, I attempted to code the solution already:
https://github.com/hashicorp/vault-k8s/pull/729

PR #703, opened 2024-11-08
feat: add agent injector telemetry

gsantos-hc:
Add Prometheus metrics to monitor the Agent Injector's performance. New metrics include a gauge of current requests being processed by the webhook, a summary of request processing times, and a count of successful and failed injections by Kubernetes namespace.

Fixes AG-005161.

### DIFF
--- a/agent-inject/handler.go
+++ b/agent-inject/handler.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault-k8s/agent-inject/agent"
@@ -24,6 +25,8 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 )
+
+const warningInitOnlyInjection = "workload uses initContainer only; consider migrating to VSO"
 
 var (
 	admissionScheme = runtime.NewScheme()
@@ -85,6 +88,14 @@ type Handler struct {
 func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) {
 	h.Log.Info("Request received", "Method", r.Method, "URL", r.URL)
 
+	// Measure request processing duration and monitor request queue
+	requestQueue.Inc()
+	requestStart := time.Now()
+	defer func() {
+		requestProcessingTime.Observe(float64(time.Since(requestStart).Milliseconds()))
+		requestQueue.Dec()
+	}()
+
 	if ct := r.Header.Get("Content-Type"); ct != "application/json" {
 		msg := fmt.Sprintf("Invalid content-type: %q", ct)
 		http.Error(w, msg, http.StatusBadRequest)
@@ -142,11 +153,20 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) {
 		msg := fmt.Sprintf("error marshalling admission response: %s", err)
 		http.Error(w, msg, http.StatusInternalServerError)
 		h.Log.Error("error on request", "Error", msg, "Code", http.StatusInternalServerError)
+		incrementInjectionFailures(admReq.Request.Namespace)
 		return
 	}
 
 	if _, err := w.Write(resp); err != nil {
 		h.Log.Error("error writing response", "Error", err)
+		incrementInjectionFailures(admReq.Request.Namespace)
+		return
+	}
+
+	if admResp.Response.Allowed {
+		incrementInjections(admReq.Request.Namespace, *admResp.Response)
+	} else {
+		incrementInjectionFailures(admReq.Request.Namespace)
 	}
 }
 
@@ -243,6 +263,13 @@ func (h *Handler) Mutate(req *admissionv1.AdmissionRequest) *admissionv1.Admissi
 	if err != nil {
 		err := fmt.Errorf("error creating patch for agent: %s", err)
 		return admissionError(req.UID, err)
+	}
+
+	// Expose when workloads are injecting initContainers only. This is useful to identify
+	// workloads that are likely good candidates for migration to Vault Secrets Operator for
+	// greater scalability and performance.
+	if agentSidecar.PrePopulateOnly {
+		resp.Warnings = append(resp.Warnings, warningInitOnlyInjection)
 	}
 
 	resp.Patch = patch
--- a/agent-inject/handler.go
+++ b/agent-inject/handler.go
@@ -26,8 +26,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-const warningInitOnlyInjection = "workload uses initContainer only; consider migrating to VSO"
-
 var (
 	admissionScheme = runtime.NewScheme()
 	deserializer    = func() runtime.Decoder {
@@ -120,7 +118,10 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var admResp admissionv1.AdmissionReview
+	var (
+		mutateResp MutateResponse
+		admResp    admissionv1.AdmissionReview
+	)
 
 	// Both v1 and v1beta1 AdmissionReview types are exactly the same, so the v1beta1 type can
 	// be decoded into the v1 type. However the runtime codec's decoder guesses which type to
@@ -137,7 +138,8 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) {
 		h.Log.Error("error on request", "Error", msg, "Code", http.StatusInternalServerError)
 		return
 	} else {
-		admResp.Response = h.Mutate(admReq.Request)
+		mutateResp = h.Mutate(admReq.Request)
+		admResp.Response = mutateResp.Resp
 	}
 
 	// Default to a v1 AdmissionReview, otherwise the API server may not recognize the request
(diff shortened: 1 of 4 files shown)


subcommand/injector/command.go
```diff
@@ -231,6 +232,7 @@ func (c *Command) Run(args []string) int {
 
 	// Registering path to expose metrics
 	if c.flagTelemetryPath != "" {
+		agentInject.MustRegisterInjectorMetrics(prometheus.DefaultRegisterer)
```
gsantos-hc, line 235: Note that this ultimately calls `registry.MustRegister`, which will [panic](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MustRegister) on failure. The behavior matches [what's done in VSO](https://github.com/hashicorp/vault-secrets-operator/blob/1273434f78c5a2561ec461715306d146cc13fe69/main.go#L366) but wanted to call it out just in case.

agent-inject/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		Name:      "request_queue_total",
+		Help:      "Total count of webhook requests in the queue",
+	})
+
+	requestProcessingTime = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: metricsSubsystem,
+		Name:      "request_processing_duration_ms",
+		Help:      "Histogram of webhook request processing times in milliseconds",
+		Buckets: []float64{
```
gsantos-hc, line 26: Went with a fixed set of buckets, similar to [what's done for VSO](https://github.com/hashicorp/vault-secrets-operator/blob/1273434f78c5a2561ec461715306d146cc13fe69/vault/client_metrics.go#L24), but we could also use something like [ExponentialBuckets](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#ExponentialBuckets)?

tvoran (commented):
This worked well when I was running it locally, just left a couple thoughts.

agent-inject/metrics.go
```diff
@@ -0,0 +1,54 @@
+package agent_inject
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	metricsNamespace      = "vault"
+	metricsSubsystem      = "sidecar_injector"
```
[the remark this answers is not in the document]
gsantos-hc: Fixed in 64edf1d6c7ee8173b35f109cc4f5321542e309a4.

agent-inject/handler.go
```diff
@@ -249,6 +264,8 @@ func (h *Handler) Mutate(req *admissionv1.AdmissionRequest) *admissionv1.Admissi
 	patchType := admissionv1.PatchTypeJSONPatch
 	resp.PatchType = &patchType
 
+	injectionsByNamespace.With(prometheus.Labels{metricsLabelNamespace: req.Namespace}).Inc()
```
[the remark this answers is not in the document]
gsantos-hc: Fixed in 632154f496aaf545920f959739469ac735e09ae5

agent-inject/metrics.go
gsantos-hc: Decided to stick with a predefined set of buckets but updated the configuration to reduce their number (reducing cardinality) and increase the highest-tier bucket to 10 seconds (from 1 second).

agent-inject/handler.go
```diff
@@ -142,11 +151,20 @@ func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) {
 		msg := fmt.Sprintf("error marshalling admission response: %s", err)
 		http.Error(w, msg, http.StatusInternalServerError)
 		h.Log.Error("error on request", "Error", msg, "Code", http.StatusInternalServerError)
+		incrementInjectionFailures(admReq.Request.Namespace)
```
tvoran, line 154: We may also want to call `incrementInjectionFailures()` for the error returns further up in this function too? But this should suffice as-is; I could even see those being a different error metric added in the future.

tvoran (dismissed):
Thanks!

gsantos-hc:
Force-pushed:
* Rebase from the latest `main`
* Add a breakdown of injection types (normal sidecar vs initContainer-only) to further help identify potential targets for migration to VSO

agent-inject/handler.go
```diff
@@ -25,6 +26,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const warningInitOnlyInjection = "workload uses initContainer only; consider migrating to VSO"
```
tvoran, line 29: Feels like this could get noisy for folks who aren't looking to move to VSO just yet? And workloads other than init-only are certainly suitable for VSO too. I think I'd prefer to leave this sort of warning out, please.

agent-inject/handler.go
```diff
@@ -245,6 +265,13 @@ func (h *Handler) Mutate(req *admissionv1.AdmissionRequest) *admissionv1.Admissi
 		return admissionError(req.UID, err)
 	}
 
+	// Expose when workloads are injecting initContainers only. This is useful to identify
+	// workloads that are likely good candidates for migration to Vault Secrets Operator for
+	// greater scalability and performance.
+	if agentSidecar.PrePopulateOnly {
+		resp.Warnings = append(resp.Warnings, warningInitOnlyInjection)
+	}
```
tvoran, line 273: Instead of adding onto the Warning that's sent back to the user here, I'd suggest changing Mutate() to return a struct that contains the AdmissionResponse along with metadata fields for what was injected, i.e. a couple of bool's for `init` and `sidecar` injections. Something like:

```go
type MutateResponse struct {
	InjectedInit    bool
	InjectedSidecar bool
	Resp            *admissionv1.AdmissionResponse
}
```

That would also make it easier for the metrics to show the other injection case of sidecar only.

Alternately we could perhaps set the injected container info in the AuditAnnotations map on the AdmissionResponse.

agent-inject/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+	admissionv1 "k8s.io/api/admission/v1"
+)
+
+const (
+	metricsNamespace         = "vault"
+	metricsSubsystem         = "agent_injector"
+	metricsLabelNamespace    = "namespace"
+	metricsLabelType         = "injection_type"
+	metricsLabelTypeDefault  = "sidecar"
+	metricsLabelTypeInitOnly = "init_only"
```
tvoran, line 16: The injector adds an init container and a sidecar container by default, so default = "sidecar" doesn't sound right to me. I'd say there are three injection types: init only, sidecar only, and both. In the future we'll hopefully support [native sidecars](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) as an injection type.

agent-inject/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+		Name:      "failed_injections_by_namespace_total",
+		Help:      "Total count of failed Agent Sidecar injections by namespace",
+	}, []string{metricsLabelNamespace})
+)
+
+func incrementInjections(namespace string, res admissionv1.AdmissionResponse) {
+	typeLabel := metricsLabelTypeDefault
+	if slices.Contains(res.Warnings, warningInitOnlyInjection) {
+		typeLabel = metricsLabelTypeInitOnly
+	}
```
tvoran, line 54: Seems like we might as well account for the sidecar-only injection type too?

agent-inject/handler.go
gsantos-hc: Sounds good. I looked into AuditAnnotations, but since those would get persisted to Kubernetes audit logs decided to go with Warnings instead.

Agreed that it's better to just change the signature of `Mutate()` and return the metadata we need for telemetry.

agent-inject/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	metricsNamespace            = "vault"
+	metricsSubsystem            = "agent_injector"
+	metricsLabelNamespace       = "namespace"
+	metricsLabelType            = "injection_type"
+	metricsLabelTypeDefault     = "default"
```
tvoran, line 12: I doubt we'll change the default anytime soon, but it might be more clear to name this label "both" or "init_and_sidecar".

tvoran (commented):
This is looking great, just a couple minor thoughts (but nothing to hold this up over).

I was also playing around with ways to test the logic in incrementInjections(), and it looks like we could use prometheus' testutil something like this: https://gist.github.com/tvoran/b6ec51e716fd10efe7bb1847def8e98a

agent-inject/handler.go
```diff
@@ -85,6 +86,14 @@ type Handler struct {
 func (h *Handler) Handle(w http.ResponseWriter, r *http.Request) {
 	h.Log.Info("Request received", "Method", r.Method, "URL", r.URL)
 
+	// Measure request processing duration and monitor request queue
+	requestQueue.Inc()
+	requestStart := time.Now()
+	defer func() {
+		requestProcessingTime.Observe(float64(time.Since(requestStart).Milliseconds()))
```
benashz, line 93: I wonder if would be worth tracking the failures as well. We do something like this in VSO here: https://github.com/hashicorp/vault-secrets-operator/blob/6ab056c22acb5dc4812620233e3b141fb9e02f9c/vault/client.go#L810

benashz (commented):
Looking good. I have included some additional feedback for your consideration. Thanks!

agent-inject/metrics.go
```diff
@@ -0,0 +1,74 @@
+package agent_inject
```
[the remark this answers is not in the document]
gsantos-hc: Fixed in a095cc4e68419b68f78fffdd0a74617de6c929ad

agent-inject/handler.go
gsantos-hc: Makes sense, and it would unblock @tvoran's feedback [here](https://github.com/hashicorp/vault-k8s/pull/703#discussion_r1847772903).

@benashz Given that it goes beyond the scope we were asked for, is it OK if I push that to a follow-on PR?

agent-inject/metrics.go
```diff
@@ -21,15 +21,14 @@ var (
...
-		Help:      "Total count of webhook requests in the queue",
+		Name:      "request_queue_length",
+		Help:      "Count of webhook requests in the injector's queue",
 	})
 
 	requestProcessingTime = prometheus.NewHistogram(prometheus.HistogramOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsSubsystem,
 		Name:      "request_processing_duration_ms",
-		Help:      "Histogram of webhook request processing times in milliseconds",
```
benashz: I think we should include the help message, just drop the extra "Histogram" bit.

PR #705, opened 2024-11-18
ci: latest five k8s kind versions

tvoran:
Test with the last five k8s versions in kind (to match the other k8s projects), and use the latest [kind node versions](https://hub.docker.com/r/kindest/node/tags).


hashiblaum (approved):

tvoran:
Thanks!

PR #716, opened 2024-12-12
fix(metrics): sidecar injection false positives

gsantos-hc:
Do not increment the `injections_by_namespace` metric when no injection is performed.


benashz (approved):
LGTM!

Issue #724, opened 2024-12-30
Add exit_after_auth as an annotation on vault-injector


plieberg:
**Is your feature request related to a problem? Please describe.**
Add the 'exit_after_auth' option as a pod annotation. Currently, the only way to use this is to use the configmap method, which adds additional, unwanted complexity to our cronjob deployments.

**Describe the solution you'd like**
I want the 'exit_after_auth' option to be available as a pod annotation so the agent will exit after authenticating and rendering the template. This is needed so cronjobs that have the vault-injector will exit properly when they have completed executing.

**Describe alternatives you've considered**
We are currently using the configmap method, but this is a deviation from our developed cronjob standard templates and requires additional work. 

**Additional context**

tvoran:
Hi @plieberg, it sounds like the existing [`agent-pre-populate-only` annotation](https://developer.hashicorp.com/vault/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-pre-populate-only) might be what you're looking for; when that is set to `true` Agent is injected only as an init container, with `exit_after_auth` set to `true`.

plieberg:
This is not needed as the `agent-pre-populate-only` annotation accomplishes the same thing.

PR #729, opened 2025-01-10
Adding support for exit after auth as an annotation

plieberg:
Fixes [#724](https://github.com/hashicorp/vault-k8s/issues/724) by adding `vault.hashicorp.com/agent-exit-after-auth`

To tell the agent to exit after auth, set annotation to true:
`vault.hashicorp.com/agent-exit-after-auth: "true"`


plieberg:
This is not needed as the `agent-pre-populate-only` annotation accomplishes the same thing.

PR #739, opened 2025-02-05
Update webhook CA bundle verification log message

ltcarbonell:
Makes log message more accurate by clarifying that this step verifies the existing CA bundle configuration rather than suggesting certificate regeneration.

Changes:
- Adjusted log level from Info to Debug
- Updated message to better reflect the actual operation


tvoran (approved):
Thanks!

PR #760, opened 2025-04-17
Add `template_config.lease_renewal_threshold` option to Agent Injector annotations

scellef:
Adds the missing Vault Agent `template_config` option to vault-k8s so `lease_renewal_threshold` can be set directly via `vault.hashicorp.com/template-lease-renewal-threshold` annotation.  Currently, the only way to set this option on injected Vault agents is to write an entire config with the parameter set, and mount it in a ConfigMap.


scellef:
Cleaning up PR hygiene, will open a new one

PR #761, opened 2025-04-22
Adding `template_config.lease_duration_threshold` to annotations

scellef:
Added missing `template_config.lease_duration_threshold` to Agent Injector annotations.  Previously, the only way to set this option in injected Vault Agents was to create an entire agent config in a ConfigMap, and reference it with the `vault.hashicorp.com/agent-configmap` annotation.  This allows for specifying the option directly via annotation `vault.hashicorp.com/template-config-lease-duration-threshold`.

### DIFF
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -295,6 +295,11 @@ const (
 	//  includes connections in the dialing, active, and idle states.
 	AnnotationTemplateConfigMaxConnectionsPerHost = "vault.hashicorp.com/template-max-connections-per-host"
 
+	// AnnotationTemplateConfigLeaseRenewalThreshold
+	// How long Vault Agent's template engine should wait for to refresh dynamic,
+	// non-renewable leases, measured as a fraction of the lease duration.
+	AnnotationTemplateConfigLeaseRenewalThreshold = "vault.hashicorp.com/template-lease-renewal-threshold"
+
 	// AnnotationAgentEnableQuit configures whether the quit endpoint is
 	// enabled in the injected agent config
 	AnnotationAgentEnableQuit = "vault.hashicorp.com/agent-enable-quit"
@@ -365,6 +370,7 @@ type AgentConfig struct {
 	ExitOnRetryFailure         bool
 	StaticSecretRenderInterval string
 	MaxConnectionsPerHost      int64
+	LeaseRenewalThreshold      float64
 	AuthMinBackoff             string
 	AuthMaxBackoff             string
 	DisableIdleConnections     string
@@ -551,6 +557,10 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 
 	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost]; !ok {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost] = strconv.FormatInt(cfg.MaxConnectionsPerHost, 10)
+	}
+
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold]; !ok {
+		pod.ObjectMeta.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold] = strconv.FormatFloat(cfg.LeaseRenewalThreshold, 'f', -1, 64)
 	}
 
 	if minBackoffString, ok := pod.ObjectMeta.Annotations[AnnotationAgentAuthMinBackoff]; ok {
@@ -863,6 +873,15 @@ func (a *Agent) templateConfigMaxConnectionsPerHost() (int64, error) {
 	}
 
 	return parseutil.ParseInt(raw)
+}
+
+func (a *Agent) templateConfigLeaseRenewalThreshold() (float64, error) {
+	raw, ok := a.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold]
+	if !ok {
+		return 0, nil
+	}
+
+	return strconv.ParseFloat(raw, 64)
 }
 
 func (a *Agent) getAutoAuthExitOnError() (bool, error) {
--- a/agent-inject/agent/annotations.go
+++ b/agent-inject/agent/annotations.go
@@ -559,10 +559,6 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost] = strconv.FormatInt(cfg.MaxConnectionsPerHost, 10)
 	}
 
-	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold]; !ok {
-		pod.ObjectMeta.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold] = strconv.FormatFloat(cfg.LeaseRenewalThreshold, 'f', -1, 64)
-	}
-
 	if minBackoffString, ok := pod.ObjectMeta.Annotations[AnnotationAgentAuthMinBackoff]; ok {
 		if minBackoffString != "" {
 			_, err := time.ParseDuration(minBackoffString)
--- a/agent-inject/agent/annotations_test.go
+++ b/agent-inject/agent/annotations_test.go
@@ -1481,30 +1481,30 @@ func TestParseTelemetryAnnotations(t *testing.T) {
 }
 
 func TestParseTemplateConfigAnnotations(t *testing.T) {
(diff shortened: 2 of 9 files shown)


agent-inject/agent/annotations.go
```diff
@@ -553,6 +559,10 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
 		pod.ObjectMeta.Annotations[AnnotationTemplateConfigMaxConnectionsPerHost] = strconv.FormatInt(cfg.MaxConnectionsPerHost, 10)
 	}
 
+	if _, ok := pod.ObjectMeta.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold]; !ok {
```
benashz, line 562: It might be good to add tests for this new annotation.

agent-inject/agent/config_test.go
```diff
@@ -675,6 +675,17 @@ func TestConfigVaultAgentTemplateConfig(t *testing.T) {
 				MaxConnectionsPerHost: 100,
 			},
 		},
+		{
```
benashz, line 678: Do we have a test for the default/unset condition?

subcommand/injector/flags_test.go
```diff
@@ -227,3 +227,31 @@ func TestCommandEnvInts(t *testing.T) {
...
 }
+
+func TestCommandEnvFloats(t *testing.T) {
+	var cmd Command
+	tests := []struct {
+		env    string
+		value  float64
+		cmdPtr *float64
+	}{
+		{env: "AGENT_INJECT_TEMPLATE_LEASE_RENEWAL_THRESHOLD", value: 0.75, cmdPtr: &cmd.flagLeaseRenewalThreshold},
```
benashz, line 238: nit: would you mind setting each member to be on its own line?

subcommand/injector/flags_test.go
```diff
@@ -227,3 +227,31 @@ func TestCommandEnvInts(t *testing.T) {
 		})
 	}
 }
+
+func TestCommandEnvFloats(t *testing.T) {
+	var cmd Command
+	tests := []struct {
+		env    string
```
benashz, line 234: nit: seems like this should be named `var` rather than env which usually associated with processes complete environment.

I see the previous tests follow this pattern. I think it is okay to make a break from it. That can be done in a separate PR, with all instances updated.

subcommand/injector/command.go
```diff
@@ -45,44 +45,45 @@ import (
...
+	flagListen                     string   // Address of Vault Server
+	flagLogLevel                   string   // Log verbosity
+	flagLogFormat                  string   // Log format
+	flagCACertFile                 string   // TLS CA Certificate to serve
+	flagCertFile                   string   // TLS Certificate to serve
+	flagKeyFile                    string   // TLS private key to serve
+	flagExitOnRetryFailure         bool     // Set template_config.exit_on_retry_failure on agent
+	flagStaticSecretRenderInterval string   // Set template_config.static_secret_render_interval on agent
+	flagMaxConnectionsPerHost      int64    // Set template_config.max_connections_per_host on agent
+	flagLeaseRenewalThreshold      float64  // Set template_config.max_connections_per_host on agent
```
benashz, line 57: doc string mismatched.

agent-inject/agent/annotations.go
```diff
@@ -553,6 +559,10 @@ func Init(pod *corev1.Pod, cfg AgentConfig) error {
...
+		pod.ObjectMeta.Annotations[AnnotationTemplateConfigLeaseRenewalThreshold] = strconv.FormatFloat(cfg.LeaseRenewalThreshold, 'f', -1, 64)
```
benashz, line 563: Is it necessary to specify a default value?

agent-inject/agent/config_test.go
scellef: Updated with unset value

agent-inject/agent/annotations.go
scellef: Added [here](https://github.com/hashicorp/vault-k8s/pull/761/files#diff-8be5beafe59d3cada9d7b742473b1b0fd003e2b4d661a167e63da1d8d121e594R1483)

subcommand/injector/flags_test.go
```diff
@@ -235,7 +235,9 @@ func TestCommandEnvFloats(t *testing.T) {
 		value  float64
 		cmdPtr *float64
 	}{
-		{env: "AGENT_INJECT_TEMPLATE_LEASE_RENEWAL_THRESHOLD", value: 0.75, cmdPtr: &cmd.flagLeaseRenewalThreshold},
+		{env: "AGENT_INJECT_TEMPLATE_LEASE_RENEWAL_THRESHOLD",
```
benashz, line 238: nit: formatting looks off. Could you move `env` down to be in line with the other members?

agent-inject/agent/annotations_test.go
```diff
@@ -1479,3 +1479,32 @@ func TestParseTelemetryAnnotations(t *testing.T) {
...
+  }
+  for name, tc := range tests {
+    t.Run(name, func(t *testing.T) {
+      pod := testPod(tc.annotations)
+      agentConfig := basicAgentConfig()
+      err := Init(pod, agentConfig)
+      require.NoError(t, err)
+      agent, err := New(pod)
+      require.NoError(t, err)
+			require.Equal(t, tc.expectedValue, agent.VaultAgentTemplateConfig.LeaseRenewalThreshold)
```
benashz, line 1507: Formatting seems strange here.

agent-inject/agent/annotations.go
benashz: It looks you are setting a default. I guess my question was could we remove this?

agent-inject/agent/annotations_test.go
scellef: Ah, my editor is goofy and I forgot to `go fmt` here.  Adjusting

agent-inject/agent/annotations.go
scellef: Whoops, you're right.  Tested and confirmed it's not needed.

subcommand/injector/flags_test.go
```diff
@@ -235,7 +235,7 @@ func TestCommandEnvFloats(t *testing.T) {
 		value  float64
 		cmdPtr *float64
 	}{
-		{env: "AGENT_INJECT_TEMPLATE_LEASE_RENEWAL_THRESHOLD",
+		{ env: "AGENT_INJECT_TEMPLATE_LEASE_RENEWAL_THRESHOLD",
```
benashz, line 238: nit: `env` should be on its own line.

agent-inject/agent/annotations_test.go
```diff
@@ -1479,3 +1479,32 @@ func TestParseTelemetryAnnotations(t *testing.T) {
...
+		annotations   map[string]string
+		expectedValue float64
+	}{
+		"lease renewal threshold set": {
+			annotations: map[string]string{
+				"vault.hashicorp.com/template-lease-renewal-threshold": "0.75",
+			},
+			expectedValue: 0.75,
+		},
+		"lease renewal threshold unset": {
```
benashz, line 1494: How do we handle invalid values?

scellef: We don't, as far as I can tell.  I followed the same pattern as the other template config options; if you set the wrong value, the k8s controller will throw errors like the following complaining about the wrong type, refusing to schedule the deployment:

```
I0429 22:17:50.755206       1 event.go:294] "Event occurred" object="default/injected-app-5b84474b6f" fieldPath="" kind="ReplicaSet" apiVersion="apps/v1" type="Warning" reason="FailedCreate" message="Error creating: admission webhook \"vault.hashicorp.com\" denied the request: error creating new agent sidecar: strconv.ParseInt: parsing \"foo\": invalid syntax"
E0429 22:17:55.877922       1 replica_set.go:544] sync "default/injected-app-5b84474b6f" failed with admission webhook "vault.hashicorp.com" denied the request: error creating new agent sidecar: strconv.ParseInt: parsing "foo": invalid syntax
```

This from specifying a string in `vault.hashicorp.com/template-config-max-connections-per-host: foo`.  The `vault.hashicorp.com/template-config-lease-renewal-threshold-duration` annotation behaves identically.  This behavior is equivalent to setting the wrong value in an agent config directly (outside of the k8s mutating webhook workflow), which will refuse to run if specified with an invalid option:

```
$ vault agent -config=/home/vault/config.json
error loading configuration from /home/vault/config.json: At -: root.template_config.lease_renewal_threshold: unknown type *ast.LiteralType
```

scellef: @benashz Added a test for invalid values in the annotation here; is this what you were looking for?

benashz (approved):

PR #747, opened 2025-02-26
Preparing v1.6.2 release

tvoran:
Updating kind, k8s versions, Vault version, and changelog.

(Previous release PR for comparison: https://github.com/hashicorp/vault-k8s/pull/719)

PR #774, opened 2025-06-24
Prepare v1.7.0 release

tvoran:
Updating changelog, kind, k8s versions, and Vault versions.

(Previous release PR for comparison: https://github.com/hashicorp/vault-k8s/pull/747)

### DIFF
--- a/deploy/injector-deployment.yaml
+++ b/deploy/injector-deployment.yaml
@@ -24,7 +24,7 @@ spec:
       serviceAccountName: "vault-injector"
       containers:
         - name: sidecar-injector
-          image: "hashicorp/vault-k8s:1.6.2"
+          image: "hashicorp/vault-k8s:1.7.0"
           imagePullPolicy: IfNotPresent
           env:
             - name: NAMESPACE


.github/workflows/tests.yaml
```diff
@@ -39,10 +39,11 @@ jobs:
       - run: echo "setting versions"
     outputs:
       # JSON encoded array of k8s versions.
-      K8S_VERSIONS: '["1.32.2", "1.31.6", "1.30.10", "1.29.14", "1.28.15"]'
-      VAULT_N: "1.18.5"
-      VAULT_N_1: "1.17.6"
-      VAULT_N_2: "1.16.3"
+      K8S_VERSIONS: '["1.33.1", "1.32.5", "1.31.9", "1.30.13", "1.29.14"]'
```
digivava, line 42: Looking at the k8s releases [page](https://kubernetes.io/releases/), it seems like there's newer versions of each of these as there was a release on 6/17/25. Should we be using those instead?

tvoran: Since we're using kind for the CI tests, we're limited to the kindest/node image versions: https://hub.docker.com/r/kindest/node/tags

digivava (dismissed):

digivava (approved):

PR #486, opened 2023-06-12
Add Environment Variables From Kubernetes Downward API

mbillow:
Use the Kubernetes downward API to expose useful runtime information to the Vault Agent container through environment variables. 

This PR adds the following environment variables:
- `NAMESPACE` (technically, this duplicates #164, but with working tests)
- `HOST_IP`
- `POD_IP`

PR #782, opened 2025-08-14
fix: option to disable base env var injection

### LINKED ISSUE #734: Hardcoded HOST_IP in agent-injector does not work on fresh clusters.
**Describe the bug**

Currently the Agent Injector Pod Template has a hardcoded env var `HOST_IP` that is used to inject the `status.hostIP` into the container.
This value does not exist on a fresh cluster without a cloud-controller-manager installed. This results in a `Init:CreateContainerConfigError` error.

This is pretty bad because vault-agent-injector cannot be used to provide secrets for the cloud-controller-manager (anymore).

**To Reproduce**
1. Create new Kubernetes cluster and dont install anything (e.g. using 1.32.1)
2. Node has no internal or external ip set because its waiting for the cloud-controller-manager
3. Install vault-agent-injector
4. Install cloud-controller-manager with vault injection annotations
5. Pod fails with `Init:CreateContainerConfigError` error because `HOST_IP` cannot be set because there is no host ip

**Expected behavior**

I think it would be good to make this optional (if its possible) or allow specification of injection templates or if the variable is not used inside the agent-injector it could be removed.

**Environment**
* Kubernetes version: kubeadm 1.32.1
* vault-k8s version: 1.18.1

**Additional context**

none

simonostendorf:
Fixes #734 

#486 introduced the environment variables `NAMESPACE`, `HOST_IP` and `POD_IP` which are not needed in the code (as I checked it) and are only there as "nice to have".

In newly created clusters where the cloud controller manager is not running (because the agent injector should be used to inject secrets into the cloud controller manager) the nodes do not have a host ip address set. So the pods cannot start because of the error `Error: host IP unknown;`.

This PR adds the ability to disable the injection of these basic env vars with a new annotation called `vault.hashicorp.com/agent-no-base-env-vars=true`.

I hope I did all relevant code changes, I am new to this project.

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

### DIFF
--- a/agent-inject/agent/agent.go
+++ b/agent-inject/agent/agent.go
@@ -188,6 +188,9 @@ type Agent struct {
 
 	// AutoAuthExitOnError is used to control if a failure in the auto_auth method will cause the agent to exit or try indefinitely (the default).
 	AutoAuthExitOnError bool
+
+	// NoBaseEnvVars controls whether the agent will inject the basic environment variables NAMESPACE, HOST_IP and POD_IP.
+	NoBaseEnvVars bool
 }
 
 type ServiceAccountTokenVolume struct {
@@ -373,6 +376,11 @@ func New(pod *corev1.Pod) (*Agent, error) {
 		iamName, iamPath = getAwsIamTokenVolume(pod)
 	}
 
+	noBaseEnvVars := false
+	if pod.Annotations[AnnotationAgentNoBaseEnvVars] == "true" {
+		noBaseEnvVars = true
+	}
+
 	agent := &Agent{
 		Annotations:               pod.Annotations,
 		ConfigMapName:             pod.Annotations[AnnotationAgentConfigMap],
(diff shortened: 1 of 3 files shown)


simonostendorf:
I found another "dirty" fix when using the following annotations:

```
  vault.hashicorp.com/agent-json-patch: |
    [
      { "op": "remove", "path": "/env/0" },
      { "op": "remove", "path": "/env/0" },
      { "op": "remove", "path": "/env/0" }
    ]
  vault.hashicorp.com/agent-init-json-patch: |
    [
      { "op": "remove", "path": "/env/0" },
      { "op": "remove", "path": "/env/0" },
      { "op": "remove", "path": "/env/0" }
    ]
```

